PR #2, opened 2022-10-13
Add initial workflows from the indico operator repository

arturo-seijas:
Add initial workflows from the indico operator repository.


merkata (dismissed):
Cool, approving.

gtrkiller (dismissed):
LGTM

.github/workflows/integration_test.yaml
```diff
@@ -0,0 +1,85 @@
...
+      - name: Build and push image
+        id: build
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          push: true
+          tags: ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ matrix.image }}:${{ github.run_id }}
+          file: ${{ matrix.image }}.Dockerfile
+
+  integration-test-microk8s:
```
jdkandersson, line 50: Is this limited to microk8s tests?

.github/workflows/integration_test.yaml
```diff
@@ -0,0 +1,85 @@
...
+          " >> /var/snap/microk8s/current/args/containerd-template.toml'
+          sudo su -c 'systemctl restart snap.microk8s.daemon-containerd.service && microk8s status --wait-ready'
+      - name: Run integration tests
+        run: |
+          echo "CHARM_NAME=$(yq '.name' metadata.yaml)" >> $GITHUB_ENV
+          args=""
+          for image_name in $(echo '${{ needs.get-images.outputs.images }}' | jq -cr '.[]'); do
+            args="${args} --${image_name}-image ${{ env.REGISTRY }}/${{ env.OWNER }}/${image_name}:${{ github.run_id }}"
+          done
+          tox -e integration -- --model testing $args
```
jdkandersson, line 79: Do you think it is worth making this a parameter so that any additional required arguments could be passed by the user?

.github/workflows/on_push.yaml
```diff
@@ -0,0 +1,53 @@
...
+        uses: canonical/charming-actions/check-libraries@1.0.3
+        with:
+          credentials: ${{ secrets.CHARMHUB_TOKEN }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+  tests:
+    name: Run tests
+    uses: ./.github/workflows/test.yaml
+  integration-tests:
+    name: Run integration tests
+    uses: ./.github/workflows/integration_test.yaml
```
jdkandersson, line 34: This kind of limits the aplicability of this workflow where, for example, the integration tests have some unique requirements (such as for Jenkins) and we can't use the standardised tests. Could this be changed to get the user to define the pre-conditions so that this can be used in more cases?

.github/workflows/publish_branch.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Get target branch
+        id: extract-branch
+        run: echo "branch=$(echo ${GITHUB_REF#refs/heads/})" >> $GITHUB_ENV
+  publish-charm:
+    name: Publish
+    uses: ./.github/workflows/publish_charm.yaml
+    needs: [ tests, integration-tests, select-branch ]
+    with:
+      channel: ${{ needs.select-branch.outputs.branch }}
+      images: ${{ needs.integration-tests.outputs.images }}
```
jdkandersson, line 38: What if there are no images? Does this still work?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,75 @@
+name: Tests
+
+on:
+  workflow_call:
+
+jobs:
+  lint-and-unit-test:
+    name: Lint and unit tests
+    runs-on: ubuntu-20.04
```
jdkandersson, line 9: This is also kind of limiting, what if the charm requires a different Ubuntu version?

jdkandersson (changes requested):
Do you think more documentation in the readme, including examples, would be helpful?

arturo-seijas:
> Do you think more documentation in the readme, including examples, would be helpful?

Yes. I think so, but I still need to evaluate [starter workflows](https://docs.github.com/en/actions/using-workflows/creating-starter-workflows-for-your-organization). I don't think it does pay off to do that right now.

.github/workflows/integration_test.yaml
arturo-seijas: That will come in handy, but I think we can add it as part of another PR. Thank you for the suggestion.

.github/workflows/integration_test.yaml
arturo-seijas: Tests are run in microk8s within that step

.github/workflows/on_push.yaml
arturo-seijas: Can you point me to those specific cases?

.github/workflows/test.yaml
arturo-seijas: We'll probably have to parametrize that. I've identified some other special cases but we'll probably have to go charm to make the appropriate changes. Do we have this scenario already?

arturo-seijas: Parametrized for both unit and integration tests. I've seen jenkins needs to run the tests on ubuntu 18. Thank you for your input

.github/workflows/integration_test.yaml
jdkandersson: I guess the question was, for the Jenkins charm (and other machine charms), the integration tests would run on lxd. I wonder if it is possible to support both?

.github/workflows/on_push.yaml
jdkandersson: Ok cool, do we need a follow up story or do you expect to be able to do it in your current story? Marking as resolved for this PR

jdkandersson (approved):

.github/workflows/integration_test.yaml
arturo-seijas: Not for now. I've made some research and the most feasible idea for now is to have two different reusable workflows. I'll work into that when the time to migrate a machine charm comes.

mthaddon (approved):
LGTM, thx

PR #4, opened 2022-10-19
Fix comment workflow

arturo-seijas:
Fix reference to comment action

### DIFF
--- a/.github/workflows/comment.yaml
+++ b/.github/workflows/comment.yaml
@@ -10,7 +10,7 @@ jobs:
     if: github.event.workflow_run.event == 'pull_request'
     steps:
       - uses: actions/checkout@v3
-      - uses: canonical/operator-workflows/.github/actions/comment_pr
+      - uses: canonical/operator-workflows/.github/actions/comment_pr@main
         with:
           artifact-name: report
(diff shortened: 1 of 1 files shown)


.github/workflows/comment.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
     if: github.event.workflow_run.event == 'pull_request'
     steps:
       - uses: actions/checkout@v3
-      - uses: canonical/operator-workflows/.github/actions/comment_pr
+      - uses: canonical/operator-workflows/.github/actions/comment_pr@main
```
weiiwang01, line 13: I think you can use relative path here, like `./.github/actions/comment_pr`
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-an-action-in-the-same-repository-as-the-workflow

arturo-seijas: I'm checking out the indico repository and the file doesn't exist in there. In the next PR I'll move this action to a new repository,  so I'd leave it as it is for now if that works for you

gregory-schiano:
LGTM

gregory-schiano (approved):
LGTM

mthaddon (approved):
LGTM

PR #9, opened 2022-10-21
Install the woke snap for inclusive naming testing

mthaddon:
We're planning to standardise on using the `woke` snap to check for inclusive naming in our projects. To do this we'll need to install the snap before running tests.


arturo-seijas (approved):
LGTM

gregory-schiano (approved):
LGTM

PR #10, opened 2022-10-21
Add support for running script before integration tests

arturo-seijas:
* Support running a script right before running the integration tests. This will allow to run cross-model tests.
* Some refactoring


gregory-schiano (approved):
LGTM
Using the provider is indeed better than splitting the steps in 2

arturo-seijas:
> LGTM Using the provider is indeed better than splitting the steps in 2

Indeed. Although it required some additional refactoring :)

mthaddon (approved):
Looks great, thanks

PR #13, opened 2022-10-24
Fix series parameter on push workflow

arturo-seijas:
Fix series parameter default value for integration tests when running the publish workflows.


mthaddon (approved):
LGTM, thx

gtrkiller (approved):
LGTM

PR #20, opened 2022-11-07
Add support for forks

arturo-seijas:
Publish images as an artifact instead of using github's container registry to be able to run integration tests involving images from forks.


gtrkiller (approved):
LGTM

merkata (approved):
Looks great! Approving.

PR #23, opened 2022-11-07
Refactor comment action and change filenames

arturo-seijas:
* Refactor comment action to include it as part of the workflow
* Change filenames


.github/workflows/comment.yaml
```diff
@@ -3,14 +3,87 @@ name: Comment on the pull request
...
     runs-on: ubuntu-22.04
     if: github.event.workflow_run.event == 'pull_request'
     steps:
       - uses: actions/checkout@v3
-      - uses: canonical/operator-workflows/.github/actions/comment_pr@main
-        with:
-          artifact-name: report
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Download test report artifact
+      uses: actions/github-script@v6.2.0
```
jdkandersson, line 17: is the indentation correct here? It looks like `uses` is not aligned with `name`

arturo-seijas: Indeed. Thx

jdkandersson (approved):

weiiwang01 (approved):

PR #24, opened 2022-11-08
Do not try to configure microk8s with LXD provider

arturo-seijas:
Only enable registry and push images if the microk8s provider has been to run the integration tests.


weiiwang01 (approved):

mthaddon (approved):
LGTM

PR #25, opened 2022-11-08
Add inclusive naming checks

arturo-seijas:
Add inclusive naming checks in the test workflow per https://github.com/canonical/Inclusive-naming

An example can be seen in https://github.com/canonical/indico-operator/pull/168

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -11,7 +11,10 @@ jobs:
       - uses: actions/checkout@v3
         with:
           repository: canonical-web-and-design/Inclusive-naming
-      - run: mv canonical-web-and-design /tmp
+      - run: |
+          pwd
+          ls
+          mv canonical-web-and-design /tmp
       - uses: actions/checkout@v3
       - name: Merge configuration files
         run: |
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -17,7 +17,7 @@ jobs:
           path: inclusive-naming
       - name: Merge configuration files
         run: |
-          yq eval-all '. as $item ireduce ({}; . *+ $item)' inclusive-naming/config.yaml current./woke.yaml > /tmp/config.yaml
+          yq eval-all '. as $item ireduce ({}; . *+ $item)' inclusive-naming/config.yml current/.woke.yaml > /tmp/config.yaml
       - name: Run inclusive naming check
         uses: canonical-web-and-design/inclusive-naming@main
         with:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -21,6 +21,7 @@ jobs:
       - name: Run inclusive naming check
         uses: canonical-web-and-design/inclusive-naming@main
         with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
           fail-on-error: "true"
           woke-args: '. -c /tmp/config.yaml'
   get-runner-image:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -11,8 +11,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           repository: canonical-web-and-design/Inclusive-naming
-      - run: mv canonical-web-and-design /tmp/canonical-web-and-design/Inclusive-naming
-          path: /tmp/inclusive-naming
+      - run: mv canonical-web-and-design /tmp
       - uses: actions/checkout@v3
       - name: Merge configuration files
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -4,6 +4,30 @@ on:
   workflow_call:
 
 jobs:
+  inclusive-naming-check:
+    name: Inclusive naming
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          repository: canonical-web-and-design/Inclusive-naming
```
mthaddon, line 13: Shouldn't this be https://github.com/canonical/Inclusive-naming ?

.github/workflows/test.yaml
```diff
@@ -4,6 +4,30 @@ on:
...
+      - name: Merge configuration files
+        run: |
+          yq eval-all '
+          (
+            . as $item ireduce ({}; . *+ $item) | .rules | unique_by(.name)
+          ) as $mergedArray | . as $item ireduce ({}; . *+ $item) | .rules = $mergedArray
+          ' .woke.yaml /tmp/config.yml > /tmp/config.yaml
+          cat /tmp/config.yaml
+      - name: Run inclusive naming check
+        uses: canonical-web-and-design/inclusive-naming@main
```
mthaddon, line 25: Same as above in terms of location.

.github/workflows/test.yaml
```diff
@@ -4,6 +4,30 @@ on:
...
+      - run: mv * /tmp
+      - uses: actions/checkout@v3
+      - name: Merge configuration files
+        run: |
+          yq eval-all '
+          (
+            . as $item ireduce ({}; . *+ $item) | .rules | unique_by(.name)
+          ) as $mergedArray | . as $item ireduce ({}; . *+ $item) | .rules = $mergedArray
+          ' .woke.yaml /tmp/config.yml > /tmp/config.yaml
+          cat /tmp/config.yaml
```
mthaddon, line 23: This is great in terms of being able to merge what's defined in that repo with what's in our location `.woke.yml`. I worry a little that this will mean that we can't run the check locally, and we'll get different results in CI than locally, but it's a minor concern. If there's a way to do this that could end up with the same result then great, but I'm not sure that's possible given this is using an externally defined github action.

mthaddon (changes requested):

.github/workflows/test.yaml
arturo-seijas: Indeed

.github/workflows/test.yaml
arturo-seijas: I agree. However, I'm not sure we'll be able to do so given that the list is different from the default snap configuration

.github/workflows/test.yaml
```diff
@@ -4,6 +4,31 @@ on:
...
+      - uses: actions/checkout@v3
+      - name: Merge configuration files
+        run: |
+          # Combine all entries and replace matching elements by
+          # .name in .rules for the ones in .woke.yaml
+          yq eval-all '
+          (
+            . as $item ireduce ({}; . *+ $item) | .rules | unique_by(.name)
+          ) as $mergedArray | . as $item ireduce ({}; . *+ $item) | .rules = $mergedArray
+          ' .woke.yaml /tmp/config.yml > /tmp/config.yaml
```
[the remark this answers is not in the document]
arturo-seijas: Good point. Missed that one. Fixed

.github/workflows/test.yaml
```diff
@@ -4,6 +4,33 @@ on:
...
+            ) as $mergedArray | . as $item ireduce ({}; . *+ $item) | .rules = $mergedArray
+            ' .woke.yaml /tmp/config.yml > /tmp/config.yaml
+          fi
+      - name: Run inclusive naming check
+        uses: canonical/inclusive-naming@main
+        with:
+          fail-on-error: "true"
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          reporter: github-pr-review
+          woke-args: '. -c /tmp/config.yaml'
```
mthaddon, line 33: Will this not find anything in `.woke.yaml` and report it as a problem? (i.e. in that file we might have an override for "whitelist", but since we're running against the current directory would it not report that as a problem?)

arturo-seijas: The file that's passed as an argument is the only one that is actually used. By default it will use the one defined in the action, which is the one that we copy to that location and override according to the contents of .woke.yaml

arturo-seijas: The extension was different in both files, so it was not overwritten, but that's fixed now

gregory-schiano (approved):
LGTM

jdkandersson (approved):

merkata (approved):
That is some nice yq expression there, nice! Approving.

mthaddon (approved):
LGTM, thx

PR #27, opened 2022-11-09
Default to ubuntu-22.04 when there's no charmcraft.yaml present

### LINKED ISSUE #26: Support repos without charmcraft.yaml
Thanks for creating these! I was going to use the test workflow for the charm documentation repo (https://github.com/canonical/upload-charm-docs). I encountered an error where, since that repo does not define a charm, there is no `charmcraft.yaml` file to figure out which Ubuntu version to use. Could this kind of repo be supported as well? For now, I just added that file and it seems to work fine.

Additionally, maybe for another issue, is there a way to support integration tests that don't need any images? Currently the workflow just stops in that case.

arturo-seijas:
Close #26

### DIFF
--- a/.github/workflows/get_runner_image.yaml
+++ b/.github/workflows/get_runner_image.yaml
@@ -17,5 +17,9 @@ jobs:
       - uses: actions/checkout@v3
       - name: Get build-on value
         run: |
-          echo "BUILD_ON=$(yq '[.bases.[].build-on.[0].name, .bases.[].build-on.[0].channel] | join("-")' charmcraft.yaml)" >> $GITHUB_ENV
+          image="ubuntu-22.04"
+          if [ -f charmcraft.yaml ]; then
+            image=$(yq '[.bases.[].build-on.[0].name, .bases.[].build-on.[0].channel] | join("-")' charmcraft.yaml)
+          fi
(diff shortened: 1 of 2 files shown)


merkata (approved):
Good!

.github/workflows/get_runner_image.yaml
```diff
@@ -17,5 +17,9 @@ jobs:
       - uses: actions/checkout@v3
       - name: Get build-on value
         run: |
-          echo "BUILD_ON=$(yq '[.bases.[].build-on.[0].name, .bases.[].build-on.[0].channel] | join("-")' charmcraft.yaml)" >> $GITHUB_ENV
+          image="ubuntu-22.04"
+          if [ -f charmcraft.yaml ]; then
+            image=$(yq '[.bases.[].build-on.[0].name, .bases.[].build-on.[0].channel] | join("-")' charmcraft.yaml)
```
gregnebu, line 22: I think you can use `yq` alternate value no ?
https://mikefarah.gitbook.io/yq/operators/alternative-default-value

gregory-schiano: Indeed

gregory-schiano (approved):
LGTM

PR #34, opened 2022-11-10
add default app name if metadata.yaml does not exist

jdkandersson:


.github/workflows/integration_test.yaml
```diff
@@ -131,7 +131,7 @@ jobs:
         run: bash ${{ inputs.pre-run-script }}
       - name: Run integration tests
         run: |
-          echo "CHARM_NAME=$(yq '.name' metadata.yaml)" >> $GITHUB_ENV
+          echo "CHARM_NAME=$(yq '.name' metadata.yaml | echo UNKNOWN)" >> $GITHUB_ENV
```
arturo-seijas, line 134: This will output also an error code. Can we check for the existence of the file instead?

arturo-seijas (changes requested):

.github/workflows/integration_test.yaml
arturo-seijas, line 134: This will output also an error code. Can we check for the existence of the file instead?

jdkandersson: Thank you, fixed

arturo-seijas (approved):
LGTM

merkata (approved):
Nice, approving.

PR #36, opened 2022-11-10
Enable ingress integration tests

jdkandersson:


.github/workflows/integration_test.yaml
```diff
@@ -126,6 +126,9 @@ jobs:
           password = \"${{ secrets.GITHUB_TOKEN }}\"
           " >> /var/snap/microk8s/current/args/containerd-template.toml'
           sudo su -c 'systemctl restart snap.microk8s.daemon-containerd.service && microk8s status --wait-ready'
+      - name: Enable microk8s ingress
+        run: |
+          sudo microk8s enable ingress
```
arturo-seijas, line 131: This will cause the pipeline to fail if the lxd provider is specified. You'll probably need to add `if: ${{ inputs.provider == 'microk8s` to prevent this step from execution under that scenario

jdkandersson: Thanks, fixed

arturo-seijas (approved):
LGTM

mthaddon (approved):
LGTM

PR #38, opened 2022-11-11
Hotfix/linters - add `runs-on` required clause

merkata:
Left out a runs-on that is required, this hotfix brings it in (using the same image as the inclusive naming above for consistency).


gtrkiller (approved):
LGTM

amandahla (approved):

PR #43, opened 2022-11-16
Fix merge of the local config with the preexisting one

arturo-seijas:


.github/workflows/test.yaml
```diff
@@ -28,7 +28,8 @@ jobs:
             (
               . as $item ireduce ({}; . *+ $item) | .rules | unique_by(.name)
             ) as $mergedArray | . as $item ireduce ({}; . *+ $item) | .rules = $mergedArray
-            ' $woke_file /tmp/config.yml > /tmp/config.yml
+            ' $woke_file /tmp/config.yml | tee /tmp/merged.yml
```
gregory-schiano, line 31: I think it's the purpose of the 'inplace' flag of yq
`yq -i ' ... ' file1 file2
Will update "file1" with the evaluated expression... but not sure it'll append both file, did you try ?

arturo-seijas: That actually updates in place the first file passed as argument, but the expression is not commutative

gregory-schiano (approved):
LGTM

mthaddon (approved):
LGTM

PR #44, opened 2022-11-17
Add an option to set up a test swift server in integration tests

weiiwang01:
Some of the charms like WordPress charm and potentially content-cache-k8s charm needs a swift server to run integration tests. Right now, the WordPress charm is using a homebrew integration workflow. By adding this option, the WordPress charm can switch to using `operator-workflows` for integration tests.


jdkandersson (approved):

gregory-schiano (approved):
LGTM

PR #46, opened 2022-11-18
feat(shellcheck): Gather shell script files

merkata:
I've combined the functionality of the previously used shellcheck action that looks for shell script files with the current shellcheck action that did not have that functionality (but properly annotates issues on PR). It is directly taken from the action with slight modifications (rather reductions) - it will scan for well-known shell files, shell scripts ending in .sh, files that are executable (via `perm /111` filter) and do not have an extension (via `! -f *.*` filter). This was tested in indico-operator prior to this PR.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -47,9 +47,60 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
+      - name: Gather files to scan
+        shell: bash
+        id: gather
+        run: |
+          declare -a filepaths
+          shebangregex="^#! */[^ ]*/(env *)?[abk]*sh"
+
+          set -f # temporarily disable globbing so that globs in inputs aren't expanded
+
+          while IFS= read -r -d '' file; do
+          filepaths+=("$file")
+          done < <(find . \
+                        -type f \
+                        '(' \
+                        -name '*.bash' \
+                        -o -name '.bashrc' \
+                        -o -name 'bashrc' \
+                        -o -name '.bash_aliases' \
+                        -o -name '.bash_completion' \
+                        -o -name '.bash_login' \
+                        -o -name '.bash_logout' \
+                        -o -name '.bash_profile' \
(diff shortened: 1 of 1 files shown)


arturo-seijas (dismissed):
LGTM

.github/workflows/test.yaml
```diff
@@ -47,9 +47,60 @@ jobs:
...
+                        -o -name 'zshrc' \
+                        -o -name '*.sh' \
+                        -o -path '*/.profile' \
+                        -o -path '*/profile' \
+                        -o -name '*.shlib' \
+                        ')' \
+                        -print0)
+
+          while IFS= read -r -d '' file; do
+          head -n1 "$file" | grep -Eqs "$shebangregex" || continue
```
gregory-schiano, line 94: I'm wondering if you couldn't use the `file` command that will probably be better at matching if it's a shell script

merkata: That's good feedback - I dug in and specifically for shell scripts there is no difference, using `file` over a shell script will try to match a shebang as listed here https://github.com/file/file/blob/ba7b3172d658e3392ff7764c48dd2039c7ef026d/magic/Magdir/commands and if the file contains valid shell code and no shebang, `file` will categorize it as regular ASCII file. The behavior with `grep` and matching against the shebang is exactly the same.

amandahla:
Hi, just to add a case where no sh file is found:

https://github.com/canonical/nginx-ingress-integrator-operator/actions/runs/3500685120/jobs/5863604550

merkata:
> Hi, just to add a case where no sh file is found:
> 
> https://github.com/canonical/nginx-ingress-integrator-operator/actions/runs/3500685120/jobs/5863604550

Yes, good catch! Now shellcheck runs conditionally if the previous step managed to collect shell script files within the repo, thanks!

jdkandersson (approved):

weiiwang01 (approved):

arturo-seijas (approved):
LGTM

PR #49, opened 2022-11-24
Add reusable workflow for testing and email notification on failure.

yhaliaw:
Example intended usage with schedule:
```
name: Scheduled tests

on:
  schedule:
     cron: "0 0 * * 1"

jobs:
  test-and-notify-failure:
    uses: canonical/operator-workflows/.github/workflows/test_and_notify_failure@main
    with:
      email-receiver: someone@example.com
    secrets: inherit
```

### DIFF
--- a/.github/workflows/test_and_notify_faliure.yaml
+++ b/.github/workflows/test_and_notify_faliure.yaml
@@ -29,21 +29,6 @@ on:
       type: string
 
 jobs:
-  lib-check:
-    # This job is the precondition of all jobs that need the Charmhub token.
-    # By disabling this job on forked repositories, we can achieve
-    # disabling the publishing to Charmhub action on forked repositories
-    # while still running tests on push events
-    if: github.repository_owner == 'canonical'
-    name: Check libraries
-    runs-on: ubuntu-22.04
-    steps:
-      - uses: actions/checkout@v3
-      - name: Check libs
-        uses: canonical/charming-actions/check-libraries@2.1.1
-        with:
-          credentials: ${{ secrets.CHARMHUB_TOKEN }}
-          github-token: ${{ secrets.GITHUB_TOKEN }}
   tests:
     name: Run tests
     uses: ./.github/workflows/test.yaml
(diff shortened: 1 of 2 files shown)


.github/workflows/test_and_notify_faliure.yaml
```diff
@@ -0,0 +1,88 @@
...
+        if: ${{ env.USERNAME == null || env.PASSWORD == null }}
+        run: |
+          echo "Missing email username and password as input for this workflow."
+          exit 1
+      - name: Notify Failure
+        uses: dawidd6/action-send-mail@v3
+        with:
+          server_address: ${{ inputs.email-server-address }}
+          server_port: ${{ inputs.email-server-port }}
+          username: ${{ env.USERNAME }}
```
jdkandersson, line 78: Is it possible to replace this with `${{ secrets.EMAIL_USERNAME }}` rather than having to pass it through the environment? It would simplify this a bit and also only provide the secrets specifically where they are needed

.github/workflows/test_and_notify_faliure.yaml
```diff
@@ -0,0 +1,88 @@
...
+    with:
+      extra-arguments: ${{ inputs.integration-test-extra-arguments }}
+      pre-run-script: ${{ inputs.integration-test-pre-run-script }}
+      provider: ${{ inputs.integration-test-provider }}
+      series: ${{ inputs.integration-test-series }}
+    secrets: inherit
+  notify-on-failure:
+    name: Notify on failure
+    needs: [integration-test]
+    if: ${{ inputs.notify && failure() }}
```
jdkandersson, line 62: Out of curiosity, where is the `inputs.notify` coming from here?

.github/workflows/test_and_notify_faliure.yaml
```diff
@@ -0,0 +1,88 @@
...
+    uses: ./.github/workflows/integration_test.yaml
+    with:
+      extra-arguments: ${{ inputs.integration-test-extra-arguments }}
+      pre-run-script: ${{ inputs.integration-test-pre-run-script }}
+      provider: ${{ inputs.integration-test-provider }}
+      series: ${{ inputs.integration-test-series }}
+    secrets: inherit
+  notify-on-failure:
+    name: Notify on failure
+    needs: [integration-test]
```
jdkandersson, line 61: Can we remove the `tests` and `lib-check` jobs since the notification only depends on `integration-test`?

jdkandersson (commented):
Thanks for putting this together!

.github/workflows/test_and_notify_faliure.yaml
yhaliaw: I have updated this.

.github/workflows/test_and_notify_faliure.yaml
yhaliaw: I have updated the code.

.github/workflows/test_and_notify_faliure.yaml
yhaliaw: I have removed lib-check as it is not needed, and add tests to dependency.

jdkandersson (dismissed):

jdkandersson:
One thing I just thought of, it might be worth adding information about the notify on failure to the readme, what do you think?

merkata:
Awesome work! Have you tested the changes with some charm repo?

jdkandersson (approved):

PR #50, opened 2022-11-28
Add module to the matrix

jdkandersson:
Adds an additional parameter `modules` so that, if the integration tests have multiple modules, they can be run in parallel. Note that currently, any integration tests in multiple modules that require charms with the same name will fail because the `model` fixture from pytest operator is module scoped and we use the fixed model `testing` which means that any attempts to redeploy a charm in another module will have failures. This parameter also speeds up test execution.

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -14,7 +14,11 @@ on:
         type: string
         default: microk8s
       series:
-        description: List of series to run the tests in JSON format, i.e. '["jammy", "focal"]'. Each element will be passed to tox as --series argument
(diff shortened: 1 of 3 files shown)


weiiwang01 (dismissed):

yhaliaw (dismissed):

merkata (dismissed):
Nice!

arturo-seijas (changes requested):
You'll want to add the new parameter also to https://github.com/canonical/operator-workflows/blob/main/.github/workflows/test_and_publish_charm.yaml

weiiwang01 (approved):

gregory-schiano (approved):
LGTM

arturo-seijas (approved):
LGTM

merkata (approved):

PR #53, opened 2022-12-06
Add setting for base for promote charm

yhaliaw:
The [release charm action from charming-actions repo](https://github.com/canonical/charming-actions/tree/main/release-charm) has base related settings. Such as `Charmcraft Base Name`, `Charmcraft Base Channel`. 
It defaults to ubuntu 20.04 amd64 if not set in input.
This caused some problem with [jenkins-agent-k8s-operator](https://github.com/canonical/jenkins-agent-k8s-operator) as the base is 22.04 and not 20.04.

This PR adds the setting for base and default to the old setting for backward compatibility.

### DIFF
--- a/.github/workflows/get_runner_image.yaml
+++ b/.github/workflows/get_runner_image.yaml
@@ -26,11 +26,11 @@ jobs:
       - name: Get build-on value
         run: |
           name="ubuntu"
-          channel="20.04"
+          channel="22.04"
           image="ubuntu-22.04"
           if [ -f charmcraft.yaml ]; then
             name=$(yq '.bases.[0].build-on.[0].name // "ubuntu"' charmcraft.yaml)
-            channel=$(yq '.bases.[0].build-on.[0].channel // '20.04'' charmcraft.yaml)
+            channel=$(yq '.bases.[0].build-on.[0].channel // '22.04'' charmcraft.yaml)
             image="$name-$channel"
           fi
           echo "BUILD_ON_NAME=$name" >> $GITHUB_ENV
--- a/.github/workflows/get_runner_image.yaml
+++ b/.github/workflows/get_runner_image.yaml
@@ -27,12 +27,11 @@ jobs:
         run: |
           name="ubuntu"
           channel="22.04"
-          image="ubuntu-22.04"
           if [ -f charmcraft.yaml ]; then
             name=$(yq '.bases.[0].build-on.[0].name // "ubuntu"' charmcraft.yaml)
             channel=$(yq '.bases.[0].build-on.[0].channel // '22.04'' charmcraft.yaml)
-            image="$name-$channel"
           fi
+          image="$name-$channel"
           echo "BUILD_ON_NAME=$name" >> $GITHUB_ENV
           echo "BUILD_ON_CHANNEL=$channel" >> $GITHUB_ENV
           echo "BUILD_ON=$image" >> $GITHUB_ENV
--- a/.github/workflows/get_runner_image.yaml
+++ b/.github/workflows/get_runner_image.yaml
@@ -3,8 +3,14 @@ name: Get runner image
 on:
   workflow_call:
     outputs:
+      name:
+        description: "Get first occurrence of build-on base name"
+        value: ${{ jobs.get-runner-image.outputs.name }}
+      channel:
+        description: "Get first occurrence of build-on base channel"
+        value: ${{ jobs.get-runner-image.outputs.channel }}
       runs-on:
-        description: "Get first occurence of build-on base as image runner"
+        description: "Get first occurrence of build-on base as image runner"
         value: ${{ jobs.get-runner-image.outputs.runs-on }}
 
 jobs:
@@ -12,6 +18,8 @@ jobs:
     name: Get runner image
     runs-on: ubuntu-22.04
     outputs:
+      name: ${{ env.BUILD_ON_NAME }}
+      channel: ${{ env.BUILD_ON_CHANNEL }}
       runs-on: ${{ env.BUILD_ON }}
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/get_runner_image.yaml
+++ b/.github/workflows/get_runner_image.yaml
@@ -17,8 +17,14 @@ jobs:
       - uses: actions/checkout@v3
       - name: Get build-on value
         run: |
+          name="ubuntu"
+          channel="20.04"
           image="ubuntu-22.04"
           if [ -f charmcraft.yaml ]; then
-            image=$(yq '[.bases.[0].build-on.[0].name, .bases.[0].build-on.[0].channel] | join("-")' charmcraft.yaml)
+            name=$(yq '.bases.[0].build-on.[0].name // "ubuntu"' charmcraft.yaml)
+            channel=$(yq '.bases.[0].build-on.[0].channel // '20.04'' charmcraft.yaml)
+            image="$name-$channel"
           fi
+          echo "BUILD_ON_NAME=$name" >> $GITHUB_ENV
+          echo "BUILD_ON_CHANNEL=$channel" >> $GITHUB_ENV
           echo "BUILD_ON=$image" >> $GITHUB_ENV
--- a/.github/workflows/promote_charm.yaml
+++ b/.github/workflows/promote_charm.yaml
@@ -24,6 +24,14 @@ jobs:
     needs: get-runner-image
     steps:
       - uses: actions/checkout@v3
+      - name: Get build-on value
+        run: |
+          name="ubuntu"
+          channel="20.04"
+          if [ -f charmcraft.yaml ]; then
+            name=$(yq '.bases.[0].build-on.[0].name' charmcraft.yaml)
+            channel=$(yq '.bases.[0].build-on.[0].channel' charmcraft.yaml)
+          fi
       - name: Release charm to channel
         uses: canonical/charming-actions/release-charm@2.1.1
         with:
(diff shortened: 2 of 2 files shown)


jdkandersson (dismissed):
This looks fine, would a reasonable alternative be to look through the charmraft.yaml file and pick the earliest/ latest version of Ubuntu? My guess would be the most defensive thing to do is pick the earliest version of Ubuntu that is supported?

yhaliaw:
> This looks fine, would a reasonable alternative be to look through the charmraft.yaml file and pick the earliest/ latest version of Ubuntu? My guess would be the most defensive thing to do is pick the earliest version of Ubuntu that is supported?

Sure, I can change it to some like how the image name and channel is read from `charmcraft.yaml` here:
https://github.com/canonical/operator-workflows/blob/main/.github/workflows/get_runner_image.yaml

This technical will make this PR a breaking change then, however it is unlike to break any repo. Hopefully.

jdkandersson:
> > This looks fine, would a reasonable alternative be to look through the charmraft.yaml file and pick the earliest/ latest version of Ubuntu? My guess would be the most defensive thing to do is pick the earliest version of Ubuntu that is supported?
> 
> Sure, I can change it to some like how the image name and channel is read from `charmcraft.yaml` here: https://github.com/canonical/operator-workflows/blob/main/.github/workflows/get_runner_image.yaml
> 
> This technical will make this PR a breaking change then, however it is unlike to break any repo. Hopefully.

What scenario would break? Maybe we just ask for feedback and also add a note in the documentation somewhere how the usage needs to be changed

yhaliaw:
> > > This looks fine, would a reasonable alternative be to look through the charmraft.yaml file and pick the earliest/ latest version of Ubuntu? My guess would be the most defensive thing to do is pick the earliest version of Ubuntu that is supported?
> > 
> > 
> > Sure, I can change it to some like how the image name and channel is read from `charmcraft.yaml` here: https://github.com/canonical/operator-workflows/blob/main/.github/workflows/get_runner_image.yaml
> > This technical will make this PR a breaking change then, however it is unlike to break any repo. Hopefully.
> 
> What scenario would break? Maybe we just ask for feedback and also add a note in the documentation somewhere how the usage needs to be changed

The old code did not pass any args related to the base, and the default of 20.04 is used:
https://github.com/canonical/charming-actions/blob/main/release-charm/action.yaml

So charm using this is not on channel 20.04.
This error should be throw: `No channel with base name ${targetBase.name}, base channel ${targetBase.channel} and base architecture ${targetBase.architecture}`

This is the entry point of the release charm:
https://github.com/canonical/charming-actions/blob/main/src/actions/release-charm/release-charm.ts

Which call this function:
https://github.com/canonical/charming-actions/blob/707c46c8c4022d6f68484553da9ce2e4f28155c5/src/services/charmcraft/charmcraft.ts#L293
At line 327.

jdkandersson:
> > > > This looks fine, would a reasonable alternative be to look through the charmraft.yaml file and pick the earliest/ latest version of Ubuntu? My guess would be the most defensive thing to do is pick the earliest version of Ubuntu that is supported?
> > > 
> > > 
> > > Sure, I can change it to some like how the image name and channel is read from `charmcraft.yaml` here: https://github.com/canonical/operator-workflows/blob/main/.github/workflows/get_runner_image.yaml
> > > This technical will make this PR a breaking change then, however it is unlike to break any repo. Hopefully.
> > 
> > 
> > What scenario would break? Maybe we just ask for feedback and also add a note in the documentation somewhere how the usage needs to be changed
> 
> The old code did not pass any args related to the base, and the default of 20.04 is used: https://github.com/canonical/charming-actions/blob/main/release-charm/action.yaml
> 
> So charm using this is not on channel 20.04. This error should be throw: `No channel with base name ${targetBase.name}, base channel ${targetBase.channel} and base architecture ${targetBase.architecture}`
> 
> This is the entry point of the release charm: https://github.com/canonical/charming-actions/blob/main/src/actions/release-charm/release-charm.ts
> 
> Which call this function: https://github.com/canonical/charming-actions/blob/707c46c8c4022d6f68484553da9ce2e4f28155c5/src/services/charmcraft/charmcraft.ts#L293 At line 327.
> 
> This will be a breaking change for repo that does not setup the base in charmcraft.yaml to 20.04. As the promote charm will now be released in a different version.
> 
> It should be fine, if the charm setup correctly.

OK, sounds good. Let's see if anyone else has any further feedback

.github/workflows/promote_charm.yaml
```diff
@@ -25,5 +37,8 @@ jobs:
         with:
           credentials: ${{ secrets.CHARMHUB_TOKEN }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          origin-channel: ${{ github.event.inputs.origin-channel }}
-          destination-channel: ${{ github.event.inputs.destination-channel }}
+          origin-channel: ${{ inputs.origin-channel }}
```
arturo-seijas, line 40: Wouldn't it be necessary to add this options here as well? https://github.com/canonical/operator-workflows/blob/main/.github/workflows/publish_charm.yaml#L57

arturo-seijas:
> > This looks fine, would a reasonable alternative be to look through the charmraft.yaml file and pick the earliest/ latest version of Ubuntu? My guess would be the most defensive thing to do is pick the earliest version of Ubuntu that is supported?
> 
> Sure, I can change it to some like how the image name and channel is read from `charmcraft.yaml` here: https://github.com/canonical/operator-workflows/blob/main/.github/workflows/get_runner_image.yaml
> 
> This technical will make this PR a breaking change then, however it is unlike to break any repo. Hopefully.

Still planning to do this?

yhaliaw:
Update the to using charmcraft.yaml build-on as the base setting.

.github/workflows/promote_charm.yaml
yhaliaw: Are you referring to the CHARMHUB_TOKEN and GITHUB_TOKEN?
They are set on line 40 and line 41.

jdkandersson: I think what arturo-seijas is saying is that setting the options added here might also be required for the part he linked to

jdkandersson (dismissed):

.github/workflows/promote_charm.yaml
yhaliaw: I have looked into that before:
https://github.com/canonical/charming-actions/tree/main/upload-charm
It does not appear to need these new arguments

weiiwang01 (dismissed):

merkata (dismissed):

.github/workflows/promote_charm.yaml
```diff
@@ -24,13 +24,21 @@ jobs:
     needs: get-runner-image
     steps:
       - uses: actions/checkout@v3
+      - name: Get build-on value
+        run: |
+          name="ubuntu"
+          channel="20.04"
+          if [ -f charmcraft.yaml ]; then
+            name=$(yq '.bases.[0].build-on.[0].name' charmcraft.yaml)
+            channel=$(yq '.bases.[0].build-on.[0].channel' charmcraft.yaml)
```
gregory-schiano, line 33: What happens if there's no match ? 
I'd rather use the alternate value syntax
https://mikefarah.gitbook.io/yq/operators/alternative-default-value

.github/workflows/promote_charm.yaml
```diff
@@ -20,10 +24,23 @@ jobs:
     needs: get-runner-image
     steps:
       - uses: actions/checkout@v3
+      - name: Get build-on value
+        run: |
+          name="ubuntu"
+          channel="20.04"
+          if [ -f charmcraft.yaml ]; then
```
arturo-seijas, line 31: Have you considered moving this logic to the `get_runner_image.yaml` file? Maybe we can parse the charmcraft.yaml file there and get all the needed options.

.github/workflows/promote_charm.yaml
yhaliaw: Thanks for the suggestion.
I have updated the code with this feature.

jdkandersson (dismissed):

.github/workflows/promote_charm.yaml
yhaliaw: This is a good idea. I updated the code.

.github/workflows/get_runner_image.yaml
```diff
@@ -3,22 +3,36 @@ name: Get runner image
...
       - name: Get build-on value
         run: |
+          name="ubuntu"
+          channel="20.04"
           image="ubuntu-22.04"
           if [ -f charmcraft.yaml ]; then
-            image=$(yq '[.bases.[0].build-on.[0].name, .bases.[0].build-on.[0].channel] | join("-")' charmcraft.yaml)
+            name=$(yq '.bases.[0].build-on.[0].name // "ubuntu"' charmcraft.yaml)
+            channel=$(yq '.bases.[0].build-on.[0].channel // '20.04'' charmcraft.yaml)
+            image="$name-$channel"
```
gregory-schiano, line 34: 2 questions here:
- If name an channel are "ubuntu" and "20.04" by default, why the default image is "ubuntu-22.04" I was expecting "ubuntu-20.04"
- Depending on the answer to the first question, if image is always name-channel, then I put `image="$name-$channel" outside the `if` statement and would remove the default image value

arturo-seijas (dismissed):
LGTM

jdkandersson (dismissed):

jdkandersson (approved):

gregory-schiano (approved):
LGTM

arturo-seijas (approved):
LGTM

PR #55, opened 2022-12-08
Add tests for Tests workflows

yanksyoon:
This PR adds workflow for testing [Tests workflow](https://github.com/canonical/operator-workflows/blob/main/.github/workflows/test.yaml) with a passing dummy tox.ini (success case).

Workflow job name follows the name in the .yaml, i.e. [test.yaml](https://github.com/canonical/operator-workflows/blob/main/.github/workflows/test.yaml) `name: Tests`

This PR does NOT contain dummy files for:
- shell scripts lint
- dockerfile lint


.github/workflows/workflow_test.yaml
```diff
@@ -0,0 +1,8 @@
+name: Workflow test
+
+on:
+  pull_request:
+
+jobs:
+  Tests:
+    uses: ./.github/workflows/test.yaml
```
jdkandersson, line 8: Final new line?

tox.ini
```diff
@@ -0,0 +1,32 @@
...
+
+[testenv:coverage-report]
+description = Create test coverage report
+output = '\
+    Name           Stmts   Miss Branch BrPart  Cover   Missing\
+    ----------------------------------------------------------\
+    sample_report     0      0     0     0    0%   0\
+    ----------------------------------------------------------\
+    TOTAL            0      0     0     0    0%'
+commands = /bin/echo -e {[testenv:coverage-report]output}
```
jdkandersson, line 32: Final new line?

tox.ini
```diff
@@ -0,0 +1,32 @@
+# Copyright 2022 Canonical Ltd.
```
jdkandersson, line 1: I'm not sure if it is possible, could this be moved into a subdirectory? E.g., using `working-directory`? The reasoning being that this is not really a tox file for the repo but more used as fixtures for the tests. Do you think it should be in a subdir?

gregory-schiano:
This looks good to me, I'll approve once the changes requested by David will be done

.github/workflows/workflow_test.yaml
```diff
@@ -0,0 +1,8 @@
+name: Workflow test
+
+on:
```
arturo-seijas, line 3: I don't think you need this file. It is possible to define multiple events triggering a pipeline so you can add `pull_request` in `./.github/workflows/test.yaml`

tox.ini
yanksyoon: Having a look weather this is possible! Thank you for the feedback :)

.github/workflows/workflow_test.yaml
yanksyoon: My initial thought process was to have separate yaml and calling it as how the end-users would use it, but second thought, i think this way does sound better :) Thanks for the feedback!

.github/workflows/test.yaml
```diff
@@ -124,7 +126,7 @@ jobs:
       - name: Install tox
         run: python3 -m pip install tox
       - name: Run tests
-        run: tox --result-json=test-result.json
+        run: tox --result-json=test-result.json -c ${{ env.TOX_INI_DIR }}
```
yanksyoon, line 129: My initial thought after some tips from @jdkandersson was to use `working-directory` as either [default run arguments](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaultsrun) or for this specific run task.
However, this wasn't applicable to following tasks that `uses` rather than `runs`.
Since currently `tox.ini` file is the only file that requires being in another directory, I think it is simpler at the moment to direct tox to target the file rather than have the workflow run in another directory and change the tasks that run afterwards.
Please let me know if you have any other comments! 😃

jdkandersson: Or even just determine it more locally rather than defining an environment variable, we can't use a parameter since we want to rely on this workflow directly

jdkandersson: This could be the reason why a separate file for the `pull_request ` might be good so that it simulates exactly how it is used by external consumers

yanksyoon: I think having an optional parameter would be better for how we will be extending these tests in the future :)
Committed changes accordingly!

.github/workflows/workflow_test.yaml
yanksyoon: Hello Arturo, an update since last comment,
I've still opted for having a separate file for the following reasons:
1. We can simulate external user's behaviors
2. Extensible inputs with different tox files / other future inputs that we might be testing(e.g. Dockerfile, shellscripts) with different scenarios (e.g. failing tox case)

:)

jdkandersson (dismissed):

.github/workflows/test.yaml
```diff
@@ -2,6 +2,11 @@ name: Tests
 
 on:
   workflow_call:
+    inputs:
+        tox-ini-dir:
```
jdkandersson, line 6: Please document this parameter in the readme

tests/tox.ini
```diff
@@ -0,0 +1,32 @@
+# Copyright 2022 Canonical Ltd.
```
jdkandersson, line 1: We will probably need folders for each workflow and multiple folders for some of the workflows with different tests

jdkandersson (approved):

gregory-schiano (approved):
LGTM

PR #61, opened 2022-12-15
Add workdir back to the path of found dockerfiles

jdkandersson:
Fixing the failing dockerlint tests


jdkandersson:
Here is a successful run on indico: https://github.com/canonical/indico-operator/actions/runs/3708756353/jobs/6286654628

.github/workflows/test.yaml
```diff
@@ -138,7 +138,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: Check for Dockerfiles
         id: dockerfiles
-        run: echo -n "found=$(find . -name "*Dockerfile" -printf "%p ")" >> $GITHUB_OUTPUT
+        run: echo -n "found=$(find . -name "*Dockerfile" -printf "${{ inputs.working-directory }}%p ")" >> $GITHUB_OUTPUT
```
gregory-schiano, line 141: Another way to produce this is to use the `working-directory` instead of `.` as lookup directory to find
`find ${{ inputs.working-directory }} -name "*Dockerfule" -printf "%p "`

jdkandersson: The command already runs in the correct working directory, what is required is to include the working directory in the path so that the action that checks the Dockerfile knows where to look. Does that make sense?

arturo-seijas (approved):
LGTM

merkata (approved):
Looks great!

.github/workflows/test.yaml
jdkandersson: Resolving for now, please let me know if there is any more feedback!

PR #64, opened 2023-01-04
add job that can be used as the required check to enforce status checks pass

jdkandersson:
The `required_status_checks` job depends on all the individual checks and is intended to be used in the branch protection settings in a repo to enforce that all the status checks pass in a PR. This reduces the maintenance effort of the required status checks in the GitHub repo which otherwise would need to be adjusted over and over as the names of tests changes. Based on: https://jdkandersson.com/2023/01/04/enforce-all-github-actions-status-checks-to-pass-in-prs/

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -175,6 +175,7 @@ jobs:
         with:
           app: ${{ env.CHARM_NAME }}
           model: testing
+  required_status_checks:
     name: Required Integration Test Status Checks
     runs-on: ubuntu-latest
     needs:
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -175,3 +175,9 @@ jobs:
         with:
           app: ${{ env.CHARM_NAME }}
           model: testing
+    name: Required Integration Test Status Checks
+    runs-on: ubuntu-latest
+    needs:
+      - integration-test
+    steps:
+      - run: echo required checks passed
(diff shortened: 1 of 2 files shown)


yanksyoon (dismissed):
Just like from the blog post! Awesome!

gregory-schiano:
Isn't there a way to add an attribute to the checks we want to put as "required" and gather the list of checks that has this attribute to built the "required" list ?
I have the feeling that we could more easily forget to add a new check in the "required_status_checks" step list, than forget to add an attribute to this new check

Not sure I'm clear in my explanations 😄

arturo-seijas (dismissed):
LGTM

jdkandersson:
> 

The only think I'm aware of that disables the "squash and merge" button until the status checks pass is this setting in GitHub:
Screenshot 2023-01-05 at 7 58 43 pm

Is that what you are referring to? The problem with it is that you have to enumerate out all of the individual checks you want to pass and then you have to maintain that list

merkata (dismissed):
cool!

jdkandersson:
Needed to merge things from main

merkata (approved):

gregory-schiano (approved):
LGTM as discussed would be great to have a flag per step, but github doesn't allow much flexibility there

jdkandersson:
> Sorry I am a little late on this problem. But I think if you just leave here empty, GitHub will require all checks to pass before merging.
> Screenshot 2023-01-06 at 23 27 30

I believe I have tried that in the past. From memory, the "squash and merge" button on PRs is greyed out in that case but it can still be pressed and the merge is accepted by GtHub. The description of the option is also "choose which status checks must pass before branches can be merged ..." which indicates that you have to chose. Would be great if I was wrong on this because then we wouldn't need to add these kind of jobs!

PR #65, opened 2023-01-05
ISD-302-make-chaos-tests-reusable-in-operator-workflows-for-charm-repos

merkata:
This feature adds the capability of reusable chaos tests. A calling repo needs to supply a set of experiments and a label for pod selection such as

```yaml
jobs:
  integration-tests:
    uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@ISD-302-make-chaos-tests-reusable-in-operator-workflows-for-charm-repos
    secrets: inherit
    with:
      chaos-enabled: true
      chaos-experiments: pod-delete
      chaos-app-label: app.kubernetes.io/name=indico
```

The other inputs have defaults that make sense in a testing environment and is tailored towards our applications, meaning that the namespace is `testing` (default of pytest-operator) and the app kind is `statefulset` (default of charmed applications).

The chaos github action does not support specifying multiple experiments, it is either "all" or a specific experiment, which hampers reusability, so I've added that option in the forked repo (I'll make a PR upstream). That's why I'm pointing to my fork currently.

A successful run is available at https://github.com/canonical/indico-operator/actions/runs/3841524657/jobs/6551337982

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -214,7 +214,7 @@ jobs:
           APP_NS: ${{ inputs.chaos-app-namespace }}
 
       - name: Run Litmus Chaos experiments
-        uses: merkata/github-chaos-actions@run-multiple-scenarios
+        uses: merkata/github-chaos-actions@feat/run-multiple-scenarios
         env:
           EXPERIMENT_NAME: ${{ inputs.chaos-experiments }}
           APP_LABEL: ${{ inputs.chaos-app-label }}
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -217,4 +217,7 @@ jobs:
         uses: merkata/github-chaos-actions@feat/run-multiple-scenarios
         env:
           EXPERIMENT_NAME: ${{ inputs.chaos-experiments }}
+          CHAOS_NAMESPACE: ${{ inputs.chaos-namespace }}
+          APP_NS: ${{ inputs.chaos-app-namespace }}
           APP_LABEL: ${{ inputs.chaos-app-label }}
+          APP_KIND: ${{ inputs.chaos-app-kind }}
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -25,6 +25,30 @@ on:
         description: Use setup-devstack-swift action to prepare a swift server for testing.
         type: boolean
         default: false
+      chaos-enabled:
+        type: boolean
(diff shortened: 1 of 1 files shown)


.github/workflows/integration_test.yaml
```diff
@@ -168,10 +192,32 @@ jobs:
...
           model: testing
+      - name: Setting up kubeconfig ENV for Github Chaos Action
+        if: ${{ inputs.chaos-enabled }}
+        run: echo ::set-env name=KUBE_CONFIG_DATA::$(sudo microk8s config | base64 -w 0)
+        env:
+          ACTIONS_ALLOW_UNSECURE_COMMANDS: true
+
+      - name: Setup Litmus
+        if: ${{ inputs.chaos-enabled }}
+        uses: merkata/github-chaos-actions@master
```
gtrkiller, line 210: Looks great, but is there a reason why we are using the actions from your personal repo instead of having them in a Canonical repo?

gtrkiller (approved):

.github/workflows/integration_test.yaml
gtrkiller: Also, it seems you have some merge conflicts to solve first

arturo-seijas (approved):
LGTM

.github/workflows/integration_test.yaml
merkata: Cool, I'm merging main now, thanks!
As for the second part - the actions repo is part of litmuschaos organization and not canonical's, are you suggesting that the fork is done within canonical organization and not within a repo under my user space?

jdkandersson:
And would it be worth creating a workflow test for this for (1) a charm and (2) a non-charm?

merkata:
> I realise this is merged, does this work for any of our projects that aren't charms (e.g., `upload-charm-docs`)?

The chaos action is indifferent to charms, it relies on an application being inside a k8s cluster and the chaos should have access to the cluster (to install the chaos engine, setup permissions for the chaos runners etc.). For an experiment to run, the inputs needed are the namespace of where the application resides, the kind of the application (deployment, statefulset) and a selector label to target the relevant objects. If `upload-charm-docs` is tested within a cluster, it can make use of chaos tests.

merkata:
> And would it be worth creating a workflow test for this for (1) a charm and (2) a non-charm?

Not sure whether it would be needed, see above. I think we can experiment with the existing workflow for charms and non-charms - if non-charms are tested inside a cluster, we can have the test structure be as the one we're using for charms - app is deployed as a statefulset inside a "testing" namespace and labeled. That way we can use a chaos action without any modification.

PR #66, opened 2023-01-06
add checks for required linters

jdkandersson:
This adds a `tox.ini` check for linters we expect to run as a team. Whilst this is not a perfect solution (e.g., the linters could be commented out), combined with a code review this should help enforce that the linters we expect to be in a repo actually are. There is more we can do in the future, e.g., deeper inspection of `tox.ini` or modifying it by adding required linters.

This PR is also intended to facilitate debate on which linters should be in the required list.


merkata (dismissed):
Nice addition!

arturo-seijas (dismissed):
LGTM

.github/workflows/test.yaml
```diff
@@ -170,6 +175,57 @@ jobs:
...
+          done
+
+          # Check cmmands
+          EXPECTED_LINT_CMDS="\
+              pydocstyle \
+              codespell \
+              flake8 \
+              isort \
+              black \
+              mypy \
```
nrobinaubertin, line 217: This is more a question than a remark: why do we include mypy in a "lint" action instead of having a separate "static analysis" one ?

.github/workflows/test.yaml
```diff
@@ -170,6 +175,57 @@ jobs:
...
+              flake8-builtins \
+              flake8-test-docs \
+              pep8-naming \
+              codespell \
+              pylint \
+              pydocstyle \
+              "
+          for EXPECTED_LINT_DEP in $EXPECTED_LINT_DEPS; do
+              DEP_REGEX="deps.*=.*\n(\s.*\n)*(\s.*$EXPECTED_LINT_DEP)"
+              python3 -c "from pathlib import Path; import re; import sys; sys.exit(1) if not re.search(r'$DEP_REGEX', Path('tox.ini').read_text()) else sys.exit(0)" || \
```
gregory-schiano, line 206: Not a big fan of mixing bash and inline python, wouldn't it be better either to do the whole step in python, or use grep here ?

.github/workflows/test.yaml
```diff
@@ -170,6 +175,57 @@ jobs:
...
+              flake8-docstrings-complete \
+              flake8-builtins \
+              flake8-test-docs \
+              pep8-naming \
+              codespell \
+              pylint \
+              pydocstyle \
+              "
+          for EXPECTED_LINT_DEP in $EXPECTED_LINT_DEPS; do
+              DEP_REGEX="deps.*=.*\n(\s.*\n)*(\s.*$EXPECTED_LINT_DEP)"
```
gregory-schiano, line 205: A comment to explain the regex would be interesting for understanding

gregory-schiano:
> This adds a `tox.ini` check for linters we expect to run as a team. Whilst this is not a perfect solution (e.g., the linters could be commented out), combined with a code review this should help enforce that the linters we expect to be in a repo actually are. There is more we can do in the future, e.g., deeper inspection of `tox.ini` or modifying it by adding required linters.
> 
> This PR is also intended to facilitate debate on which linters should be in the required list.

Shouldn't we wait for the discussion on the Jira ticket to land ?
I'm not fan of parsing the tox.ini... would rather parse the tox stdout/stderr as mention in the Jira ticket

.github/workflows/test.yaml
jdkandersson: Yes that is a good point, I'll add that Monday, the regex basically looks for a line starting with `dep`, then allows for any number of lines that have whitespace at the start followed by a line that has whitespace and then the linter

.github/workflows/test.yaml
jdkandersson: There is a static analysis section, I wonder what the difference between a linter and static analysis would be?

.github/workflows/test.yaml
jdkandersson: Now that it is based on the command output, `grep` is used

jdkandersson:
> > This adds a `tox.ini` check for linters we expect to run as a team. Whilst this is not a perfect solution (e.g., the linters could be commented out), combined with a code review this should help enforce that the linters we expect to be in a repo actually are. There is more we can do in the future, e.g., deeper inspection of `tox.ini` or modifying it by adding required linters.
> > This PR is also intended to facilitate debate on which linters should be in the required list.
> 
> Shouldn't we wait for the discussion on the Jira ticket to land ? I'm not fan of parsing the tox.ini... would rather parse the tox stdout/stderr as mention in the Jira ticket

Changed to be based on stdout

amandahla (approved):

gregory-schiano (approved):
Cool great work !

.github/workflows/test.yaml
nrobinaubertin: In my mind, a linter was more about codestyle. But that means that a linter is a static analysis tool so nevermind

PR #68, opened 2023-01-10
feat(chaos): Add experiment duration

merkata:
This PR makes the TOTAL_CHAOS_DURATION configurable as an input - that is needed for charms like Discourse that take longer to recover from a chaos experiment such as a `pod delete` and the default of 60 seconds is not enough.


merkata:
Once this gets merged I'll open a complementing PR in Discourse charm with an increased duration.

arturo-seijas (approved):
LGTM

weiiwang01 (approved):

PR #70, opened 2023-01-11
feat(ISD-374): Add ZAP GitHub Action Full Scan

amandahla:
- Remove set-env
- Update README.md
- Add following new inputs:
```yaml
     zap-dast-enabled:
        type: boolean
        description: Whether ZAP testing is enabled
        default: false
      zap-auth-header:
        description: If this is defined then its value will be added as a header to all of the ZAP requests
        type: string
      zap-auth-header-value:
        description:  If this is defined then its value will be used as the header name to all of the ZAP requests
        type: string
      zap-target:
        description:  If this is not set, the unit IP address will be used as ZAP target
        type: string
      zap-target-protocol:
        description:  ZAP target protocol
        type: string
        default: "http"
      zap-target-port:
        description:  ZAP target port
        type: string
        default: 80
      zap-before-command:
        description: Command to run before ZAP testing
        type: string
      zap-options:
        description: Options to be used by ZAP
        type: string
        default: "-T 60"
```

Note: Since the ZAP testing runs together with the Integration test, the option fail_action is set to False to prevent the entire action to fail if vulnerabilities are found.

References:
https://www.zaproxy.org/docs/docker/full-scan/
https://github.com/marketplace/actions/owasp-zap-full-scan

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -77,10 +77,13 @@ on:
       zap-before-command:
         description: Command to run before ZAP testing
         type: string
-      zap-options:
+      zap-cmd-options:
         description: Options to be used by ZAP
         type: string
         default: "-T 60"
+      zap-rules-file-name:
+        description: Rules file to ignore any alerts from the ZAP scan
+        type: string
     outputs:
       images:
         description: Pushed docker images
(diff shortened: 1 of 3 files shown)


jdkandersson:
Would it be worth having a description on the purpose of the zap tests in the readme? Also, would it be worth including an integration test that both doesn't and does enable the zap testing against some dummy application? That way we have an example for how to do it in the repo

arturo-seijas (commented):
Could you add a sample run to the description when available?

OWASPZAP.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Default
+
+```yaml
+jobs:
+  integration-tests:
+    uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
+    secrets: inherit
+    with:
+      zap-dast-enabled: true
```
jdkandersson, line 34: If it is default enabled, what target does it use for the test?

.github/workflows/integration_test.yaml
```diff
@@ -75,6 +75,37 @@ on:
           Retry is chaos-duration/chaos-delay times and each retry
           sleeps chaos-delay
         default: 90
+      zap-dast-enabled:
```
arturo-seijas, line 78: Would you mind ordering these alphabetically?

arturo-seijas (approved):
LGTM

OWASPZAP.md
amandahla: it was only in the README.md file ("If this is not set, the unit IP address will be used as ZAP target"). I added to the OWASPZAP.md too now.

weiiwang01 (approved):

PR #73, opened 2023-01-12
Feat/configurable chaos duration

merkata:
This feature enables charms that take longer to recover (like discourse) to pass an experiment such as pod-delete successfully. It relies on three variables that are set:

CHAOS_INTERVAL: when to attempt another run of the experiment
DURATION: used for retries (see example below)
DELAY: sleep time in seconds

When checking the status of an experiment, the flow is that the engine runner is checked (starts a pod that controls the experiment), the pod runner is checked (the started pod in control of the experiment) and finally the application pod targeted by the experiment is checked. All three of them are checked with the same retry scheme, which is:

For DURATION/DELAY times, check the status, sleep DELAY seconds and retry if needed.

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -51,8 +51,24 @@ on:
         default: statefulset
       chaos-duration:
         type: string
-        description: Duration of the chaos experiment
+        description: |
+          Duration of the chaos experiment (added to 600s for go test timeout)
         default: 60
+      chaos-interval:
+        type: string
+        description: Interval to attempt re-run of an experiment
+        default: 30
+      chaos-status-duration:
+        type: string
+        description: |
+          Duration used for status check retries
+          Retry is chaos-duration/chaos-delay times and each retry
+          sleeps chaos-delay
+        default: 90
+      chaos-status-delay:
+        type: string
+        description: Delay is the wait(sleep) time on every status retry
+        default: 5
     outputs:
       images:
         description: Pushed docker images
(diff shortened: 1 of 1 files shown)


.github/workflows/integration_test.yaml
```diff
@@ -51,8 +51,24 @@ on:
...
+        description: Interval to attempt re-run of an experiment
+        default: 30
+      chaos-status-duration:
+        type: string
+        description: |
+          Duration used for status check retries
+          Retry is chaos-duration/chaos-delay times and each retry
+          sleeps chaos-delay
+        default: 90
+      chaos-status-delay:
```
arturo-seijas, line 68: Would you mind ordering the inputs alphabetically?

.github/workflows/integration_test.yaml
```diff
@@ -226,6 +243,9 @@ jobs:
           APP_LABEL: ${{ inputs.chaos-app-label }}
           APP_KIND: ${{ inputs.chaos-app-kind }}
           TOTAL_CHAOS_DURATION: ${{ inputs.chaos-duration }}
+          CHAOS_INTERVAL: ${{ inputs.chaos-interval }}
```
arturo-seijas, line 246: Same here for the env vars

.github/workflows/integration_test.yaml
merkata: Good point and adds to readability, I'll order them - thanks!

arturo-seijas (approved):
LGTM

amandahla (approved):

PR #78, opened 2023-01-18
feat(ISD-375): Add Trivy GitHub Action Full Scan 

amandahla:
- Two options for scanning: fs and image
- Documentation
- Change K6 action to be run before scannings
- Change Zap action to be baseline instead of full

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -202,6 +202,7 @@ jobs:
           name: ${{ matrix.image }}
           path: /tmp/${{ matrix.image }}.tar
       - name: Run Github Trivy Image Action
+        if: ${{ github.event.pull_request.head.repo.fork && inputs.trivy-image-enabled }}
         uses: aquasecurity/trivy-action@master
         with:
           input: /tmp/${{ matrix.image }}.tar
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -182,6 +182,14 @@ jobs:
           push: true
           tags: ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ matrix.image }}:${{ github.run_id }}
           file: ${{ inputs.working-directory }}/${{ matrix.image }}.Dockerfile
+      - name: Run Github Trivy Image Action
+        uses: aquasecurity/trivy-action@master
+        with:
+          image-ref: ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ matrix.image }}:${{ github.run_id }}
+          trivy-config: ${{ inputs.trivy-image-config }}
+        env:
+          TRIVY_USERNAME: ${{ github.actor }}
+          TRIVY_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
       # GitHub doesn't currently support pushing images in the Docker registry when opening a PR from a fork
       # so an artifact is published instead
       - name: Set up Docker Buildx
@@ -201,12 +209,6 @@ jobs:
         with:
           name: ${{ matrix.image }}
           path: /tmp/${{ matrix.image }}.tar
-      - name: Run Github Trivy Image Action
-        if: ${{ github.event.pull_request.head.repo.fork }}
-        uses: aquasecurity/trivy-action@master
-        with:
-          input: /tmp/${{ matrix.image }}.tar
-          trivy-config: ${{ inputs.trivy-image-config }}
 
   integration-test:
     name: Integration tests
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -94,10 +94,6 @@ on:
         type: string
         description: Target directory to do the Trivy testing
         default: "."
-      trivy-image-enabled:
-        type: boolean
-        description: Whether Trivy testing of type image is enabled
-        default: false
       trivy-image-config:
         type: string
         description: Trivy YAML configuration for image testing that is checked in as part of the repo
@@ -206,7 +202,6 @@ jobs:
           name: ${{ matrix.image }}
           path: /tmp/${{ matrix.image }}.tar
       - name: Run Github Trivy Image Action
-        if: ${{ inputs.trivy-image-enabled }}
         uses: aquasecurity/trivy-action@master
         with:
           input: /tmp/${{ matrix.image }}.tar
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -202,7 +202,7 @@ jobs:
           name: ${{ matrix.image }}
           path: /tmp/${{ matrix.image }}.tar
       - name: Run Github Trivy Image Action
-        if: ${{ github.event.pull_request.head.repo.fork && inputs.trivy-image-enabled }}
+        if: ${{ github.event.pull_request.head.repo.fork }}
         uses: aquasecurity/trivy-action@master
         with:
           input: /tmp/${{ matrix.image }}.tar
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -93,6 +93,7 @@ on:
       trivy-fs-ref:
         type: string
         description: Target directory to do the Trivy testing
+        default: "."
       trivy-image-config:
         type: string
         description: Trivy YAML configuration for image testing that is checked in as part of the repo
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -97,10 +97,6 @@ on:
       trivy-image-config:
         type: string
         description: Trivy YAML configuration for image testing that is checked in as part of the repo
-      trivy-image-enabled:
-        type: boolean
-        description: Whether Trivy testing of type image is enabled
-        default: false
       trivy-image-ref:
         type: string
         description: Image reference, e.g. alpine:3.10.2
@@ -335,7 +331,7 @@ jobs:
           scan-ref: ${{ inputs.trivy-fs-ref }}
           trivy-config: ${{ inputs.trivy-fs-config }}
       - name: Run Github Trivy Image Action
-        if: ${{ inputs.trivy-image-enabled && inputs.trivy-image-ref != '' }}
+        if: ${{ inputs.trivy-image-ref != '' }}
         uses: aquasecurity/trivy-action@master
         with:
           scan-type: 'image'
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -94,12 +94,13 @@ on:
         type: string
         description: Target directory to do the Trivy testing
         default: "."
+      trivy-image-enabled:
+        type: boolean
+        description: Whether Trivy testing of type image is enabled
+        default: false
       trivy-image-config:
         type: string
         description: Trivy YAML configuration for image testing that is checked in as part of the repo
-      trivy-image-ref:
-        type: string
-        description: Image reference, e.g. alpine:3.10.2
       working-directory:
         type: string
         description: The working directory for jobs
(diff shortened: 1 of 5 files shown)


yanksyoon (dismissed):
LGTM! Very cool!

OWASPZAP.md
```diff
@@ -66,42 +67,82 @@ jobs:
...
-headers = dict({"Host": "indico.local"})
 
-def sendingRequest(msg, initiator, helper):
-    for item in list(headers):
-      msg.getRequestHeader().setHeader(item, headers[item])
-
-def responseReceived(msg, initiator, helper):
-    pass;
+```python
+# pylint: disable=missing-module-docstring,invalid-name,unused-argument
```
jdkandersson, line 116: Are these needed in the documentation?

jdkandersson (commented):
Would it also be worth adding an integration test here?

.github/workflows/integration_test.yaml
```diff
@@ -300,6 +321,26 @@ jobs:
           EXPERIMENT_NAME: ${{ inputs.chaos-experiments }}
           TOTAL_CHAOS_DURATION: ${{ inputs.chaos-duration }}
           KUBE_CONFIG_DATA: ${{ env.KUBE_CONFIG_DATA }}
+      - name: Install k6s
+        if: ${{ inputs.load-test-enabled }}
+        run: sudo snap install k6
+      - name: Run k6 load tests
+        if: ${{ inputs.load-test-enabled }}
+        run: k6 run load_tests/load-test.js ${{ inputs.load-test-run-args }}
+      - name: Run Github Trivy FS Action
```
arturo-seijas, line 330: Any reason why trivy is run before k6s (when the last one runs)?

TRIVY.md
```diff
@@ -0,0 +1,60 @@
...
+
+The result will be available in the integration test output.
+
+## Warning
+
+- Environment variable has preference over configuration set by ``trivy-fs-config`` or ``trivy-image-config`` parameters.
+
+## How to use
+
+If there is no need for customization, the test can be enabled by setting the parameter ``trivy-fs-enabled`` or ``trivy-image-enabled`` to true.
```
arturo-seijas, line 17: This is not what I see in the code. Both are required:
```
- name: Run Github Trivy Image Action
        if: ${{ inputs.trivy-image-enabled && inputs.trivy-image-ref != '' }}
```

.github/workflows/integration_test.yaml
```diff
@@ -300,6 +321,26 @@ jobs:
...
+        if: ${{ inputs.trivy-fs-enabled }}
+        uses: aquasecurity/trivy-action@master
+        with:
+          scan-type: 'fs'
+          scan-ref: ${{ inputs.trivy-fs-ref }}
+          trivy-config: ${{ inputs.trivy-fs-config }}
+      - name: Run Github Trivy Image Action
+        if: ${{ inputs.trivy-image-enabled && inputs.trivy-image-ref != '' }}
```
arturo-seijas, line 338: Is this the desired behaviour? If so, can we get rid of the param `trivy-image-enabled`?

README.md
```diff
@@ -39,6 +39,12 @@ The following workflows are available:
 | provider | string | microk8s | Actions operator provider as defined [here](https://github.com/charmed-kubernetes/actions-operator#usage) |
 | series | string | '[""]' | List of series to run the tests in JSON format, i.e. '["jammy", "focal"]'. Each element will be passed to pytest through tox as --series argument |
 | setup-devstack-swift | bool | false | Use setup-devstack-swift action to prepare a swift server for testing. |
+| trivy-fs-config | string | "" | Trivy YAML configuration for fs type |
+| trivy-fs-enabled | boolean | false | Whether Trivy testing of type fs is enabled |
+| trivy-fs-ref | string | "." | Whether Trivy testing of type fs is enabled |
```
arturo-seijas, line 44: The description of this param is the same as the one above

README.md
```diff
@@ -39,6 +39,12 @@ The following workflows are available:
...
+| trivy-image-config | string | "" | Trivy YAML configuration for image type |
+| trivy-image-enabled | boolean | false | Whether Trivy testing of type image is enabled |
+| trivy-image-ref | string | "" | Whether Trivy testing of type image is enabled |
```
arturo-seijas, line 47: Same here

TRIVY.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Description
+
+By enabling Trivy in the integration test, the [Trivy GitHub action](https://github.com/aquasecurity/trivy-action) will run a scan in the repository (fs type) or a Docker image (image type) to find vulnerabilities.
+
+The result will be available in the integration test output.
+
+## Warning
+
+- Environment variable has preference over configuration set by ``trivy-fs-config`` or ``trivy-image-config`` parameters.
```
arturo-seijas, line 13: s/variable has/variables have

arturo-seijas (commented):
Can you link a test run, please?

.github/workflows/integration_test.yaml
amandahla: I changed k6 to run before ZAP/Trivy scanning actually because ZAP full scan was simulating attacks and impacting Indico performance. So when k6 ran after the scanning, the http_req_duration was not as expected.
Example:
https://github.com/canonical/indico-operator/actions/runs/3952422613/jobs/6767561531

I changed for the baseline scan that is faster but even so left k6 first.

.github/workflows/integration_test.yaml
amandahla: You're right, there is no need of trivy-image-enabled, removing.

amandahla:
> Can you link a test run, please?

Sure.
https://github.com/canonical/indico-operator/actions/runs/3953219831/jobs/6779700541

Nothing was found.

OWASPZAP.md
amandahla: No, removed.

.github/workflows/integration_test.yaml
arturo-seijas: Thanks for clarifying

.github/workflows/integration_test.yaml
```diff
@@ -328,14 +324,15 @@ jobs:
...
+        id: trivy-fs-action
         if: ${{ inputs.trivy-fs-enabled }}
         uses: aquasecurity/trivy-action@master
         with:
           scan-type: 'fs'
           scan-ref: ${{ inputs.trivy-fs-ref }}
           trivy-config: ${{ inputs.trivy-fs-config }}
       - name: Run Github Trivy Image Action
-        if: ${{ inputs.trivy-image-enabled && inputs.trivy-image-ref != '' }}
+        if: ${{ inputs.trivy-image-ref != '' }}
```
arturo-seijas, line 335: Would it be better to run the image scan as the last step of the `build-images` job? That way there would be no need for parametrizing the image name

amandahla:
> Would it also be worth adding an integration test here?

I tried but the "needs" context doesn't have a step outcome. My suggestion would be to work on that in a different story.

.github/workflows/integration_test.yaml
amandahla: Unfortunately, this parameter doesn't support multiple images, so I left it to be set.
There is a solution here but without using the action.
https://github.com/aquasecurity/trivy-action/issues/130

arturo-seijas: The thing is that the `build-images` jobs is running in a matrix, so we would effectively be running this action once per image built.

amandahla: Nice! Thanks

.github/workflows/integration_test.yaml
```diff
@@ -94,12 +94,13 @@ on:
         type: string
         description: Target directory to do the Trivy testing
         default: "."
+      trivy-image-enabled:
```
arturo-seijas, line 97: Maybe we can run it always? Is there any reason to skip it? It is my understanding that we can provide config to ignore issues

amandahla: Hm...Since all the other tests has the option to enable/disable, I left this one too. In case you want to disable everything just do the integration test.

arturo-seijas: I was thinking that it would make forcing compliance easier if we don't allow disabling it. Same approach as the linters

amandahla: You convinced me :)

arturo-seijas (dismissed):
LGTM

.github/workflows/integration_test.yaml
```diff
@@ -187,6 +201,12 @@ jobs:
         with:
           name: ${{ matrix.image }}
           path: /tmp/${{ matrix.image }}.tar
+      - name: Run Github Trivy Image Action
+        if: ${{ github.event.pull_request.head.repo.fork && inputs.trivy-image-enabled }}
```
arturo-seijas, line 205: We don't have this param anymore

TRIVY.md
```diff
@@ -0,0 +1,58 @@
...
+
+The result will be available in the integration test output.
+
+## Warning
+
+- Environment variables have preference over configuration set by ``trivy-fs-config`` or ``trivy-image-config`` parameters.
+
+## How to use
+
+If there is no need for customization, the test can be enabled by setting the parameter ``trivy-fs-enabled`` or ``trivy-image-enabled`` to true.
```
arturo-seijas, line 17: `trivy-image-enabled`doesn't exist anymore

.github/workflows/integration_test.yaml
amandahla: I miss a Head Banging Against Wall emoji here now...

.github/workflows/integration_test.yaml
```diff
@@ -187,6 +201,12 @@ jobs:
         with:
           name: ${{ matrix.image }}
           path: /tmp/${{ matrix.image }}.tar
+      - name: Run Github Trivy Image Action
+        if: ${{ github.event.pull_request.head.repo.fork }}
```
arturo-seijas, line 205: Is there any reason for skipping the image analysis when running from a fork?

amandahla: I'll have to check if it works accessing the registry. Would be something like this:

```
      - name: Log in to the Container registry
        if: ${{ !github.event.pull_request.head.repo.fork }}
        uses: docker/login-action@v2
        with:
          registry: ${{ env.REGISTRY }}
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}
      - name: Build and push image
        uses: docker/build-push-action@v3
        if: ${{ !github.event.pull_request.head.repo.fork }}
        with:
          context: .
          push: true
          tags: ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ matrix.image }}:${{ github.run_id }}
          file: ${{ inputs.working-directory }}/${{ matrix.image }}.Dockerfile
      - name: Run Github Trivy Image Action
        uses: aquasecurity/trivy-action@master
        with:
          image-ref: ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ matrix.image }}:${{ github.run_id }}
          trivy-config: ${{ inputs.trivy-image-config }}
        env:
          TRIVY_USERNAME: ${{ github.actor }}
          TRIVY_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
```

arturo-seijas: The tarball should be present regardless of the registry. Later it is pushed to the GH registry or to the microk8s one, but at this point that shouldn't make any difference

arturo-seijas: For forks you'd probably have to use the tarball generated in the "Build image as tarball" step

amandahla: Maybe I misunderstood. 
```
      - name: Build image as tarball
        uses: docker/build-push-action@v3
```
I thought that the tarball would be built only in this case of this if. I changed that because the tar was not found here:
https://github.com/canonical/operator-workflows/actions/runs/3960365804/jobs/6784446185#step:9:16

amandahla: I'll add there too.

arturo-seijas (approved):
LGTM

gtrkiller (approved):
LGTM

jdkandersson:
Did we end up adding any integration tests for this?

PR #83, opened 2023-01-24
Copy kube config file

arturo-seijas:
Copy kube config file in ./kube/config

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -244,6 +244,8 @@ jobs:
         uses: charmed-kubernetes/actions-operator@main
         with:
           provider: ${{ inputs.provider }}
+      - name: Copy kube-config file
+        run: sudo microk8s config > ~/.kube/config
       - name: Enable microk8s registry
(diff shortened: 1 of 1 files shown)


jdkandersson:
Could we have an integration test that checks for this failure please?

.github/workflows/integration_test.yaml
```diff
@@ -244,6 +244,8 @@ jobs:
         uses: charmed-kubernetes/actions-operator@main
         with:
           provider: ${{ inputs.provider }}
+      - name: Copy kube-config file
+        run: sudo microk8s config > ~/.kube/config
```
jdkandersson, line 248: Is it worth adding an integration test that checks for this failure?

jdkandersson (changes requested):

.github/workflows/integration_test.yaml
arturo-seijas: That code is already exercised

arturo-seijas:
Closing this Pr since it might not be needed for now.

PR #84, opened 2023-01-24
disable problematic job

jdkandersson:
The `release-charm-libs` job in `test_and_publish_charm.yaml` is causing these kind of failures: https://github.com/canonical/indico-operator/actions/runs/4000676362

An issue has been filed to address the issue: https://github.com/canonical/charming-actions/issues/82

Until then, this PR disables that job


merkata (approved):
Thanks for spotting, that's an interesting bug.

arturo-seijas (changes requested):
I think this is a preexisting issue that appears when manually running the workflow. Have you tried triggering it from a push event?

arturo-seijas (approved):
LGTM

PR #88, opened 2023-01-31
ci(action): Add Jira sync

merkata:
The Jira sync used at https://github.com/beliaev-maksim/github-to-jira-automation/blob/master/.github/workflows/issues_to_jira.yaml is not meant to be reused (missing on.workflow_call), so I'm placing it in our workflows and under Canonical. Also reformatted a bit.

### DIFF
--- a/.github/workflows/jira.yaml
+++ b/.github/workflows/jira.yaml
@@ -27,7 +27,7 @@ jobs:
 
           # send JIRA request
 
-          # Canonical discourages discussions in JIRA on pulic project issues.
+          # Canonical discourages discussions in JIRA on public project issues.
           # It is NOT recommended to put description in JIRA in order to
           # push conversation to GitHub. If strongly required,
           # replace '--arg body ""' with '--arg body "$ISSUE_DESCRIPTION"'
--- a/.github/workflows/jira.yaml
+++ b/.github/workflows/jira.yaml
@@ -1,0 +1,44 @@
+# this workflow requires to provide JIRA webhook URL via JIRA_URL GitHub Secret
+# read more: https://support.atlassian.com/cloud-automation/docs/jira-automation-triggers/#Automationtriggers-Incomingwebhook
+# original code source: https://github.com/beliaev-maksim/github-to-jira-automation
+
+name: Issues to JIRA
+
+on:
+  workflow_call:
+
+jobs:
+  update:
+    name: Update Issue
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create JIRA ticket
+        env:
+          # put into env vars to properly escape special bash chars
+          ISSUE_TITLE: ${{ github.event.issue.title }}
+          ISSUE_DESCRIPTION: ${{ github.event.issue.body }}
+        run: |
+          # Determine Bug/Enhancement
+          if ${{ contains(github.event.*.labels.*.name, 'Bug') }}; then
+            type=bug
+          else
+            type=story
+          fi
+
+          # send JIRA request
+
+          # Canonical discourages discussions in JIRA on pulic project issues.
+          # It is NOT recommended to put description in JIRA in order to
+          # push conversation to GitHub. If strongly required,
+          # replace '--arg body ""' with '--arg body "$ISSUE_DESCRIPTION"'
+
+          data=$( jq -n \
+                  --arg title "$ISSUE_TITLE" \
+                  --arg url '${{ github.event.issue.html_url }}' \
+                  --arg submitter '${{ github.event.issue.user.login }}' \
+                  --arg body "" \
+                  --arg type "$type" \
+                  --arg action '${{ github.event.action }}' \
+                  '{title: $title, url: $url, submitter: $submitter, body: $body, type: $type, action: $action}' )
+
+          curl -X POST -H 'Content-type: application/json' --data "${data}" "${{ secrets.JIRA_URL }}"
--- a/.github/workflows/jira.yaml
+++ b/.github/workflows/jira.yaml
@@ -36,7 +36,7 @@ jobs:
                   --arg title "$ISSUE_TITLE" \
                   --arg url '${{ github.event.issue.html_url }}' \
                   --arg submitter '${{ github.event.issue.user.login }}' \
-                  --arg body "" \
+                  --arg body "$ISSUE_DESCRIPTION" \
                   --arg type "$type" \
                   --arg action '${{ github.event.action }}' \
                   '{title: $title, url: $url, submitter: $submitter, body: $body, type: $type, action: $action}' )


arturo-seijas (dismissed):
LGTM

.github/workflows/jira.yaml
```diff
@@ -0,0 +1,44 @@
...
+          # Determine Bug/Enhancement
+          if ${{ contains(github.event.*.labels.*.name, 'Bug') }}; then
+            type=bug
+          else
+            type=story
+          fi
+
+          # send JIRA request
+
+          # Canonical discourages discussions in JIRA on pulic project issues.
```
mthaddon, line 30: s/pulic/public/ ?

merkata: Oh nice catch! That's in the original repo as well :) Replacing.

.github/workflows/jira.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+          # Canonical discourages discussions in JIRA on public project issues.
+          # It is NOT recommended to put description in JIRA in order to
+          # push conversation to GitHub. If strongly required,
+          # replace '--arg body ""' with '--arg body "$ISSUE_DESCRIPTION"'
+
+          data=$( jq -n \
+                  --arg title "$ISSUE_TITLE" \
+                  --arg url '${{ github.event.issue.html_url }}' \
+                  --arg submitter '${{ github.event.issue.user.login }}' \
```
gregory-schiano, line 38: Will this be ok if the submitter is someone external to Canonical and therefore won't have an account on Jira ?

.github/workflows/jira.yaml
```diff
@@ -0,0 +1,44 @@
...
+                  --arg body "" \
+                  --arg type "$type" \
+                  --arg action '${{ github.event.action }}' \
+                  '{title: $title, url: $url, submitter: $submitter, body: $body, type: $type, action: $action}' )
+
+          curl -X POST -H 'Content-type: application/json' --data "${data}" "${{ secrets.JIRA_URL }}"
```
gregory-schiano, line 44: I'm surprised, isn't there any API key secret or authentication method required to post to Jira ?
How does this work ?

merkata: The API key is part of the URL, that's why I created it as a secret. It is generated as a webhook URL by Jira in the Automation section.

.github/workflows/jira.yaml
merkata: I didn't even think about that, thanks for raising this! I did a manual curl with a bogus user - the Reporter of the issue in Jira is "Automation for Jira", the `submitter` is used only in the description field when creating the story, together with a link to the original issue. It is for informational purposes, so it works with all submitters.

gregory-schiano: Yeap so it uses the user whose API token is attached to

jdkandersson:
Thanks for adding this. Here comes the question, any tests that exercise this? How do we know it works?

merkata:
> Thanks for adding this. Here comes the question, any tests that exercise this? How do we know it works? In this case it is a bit more grey since we wouldn't run this on CI/CD. Although, if we want to use it to rely on getting notifications that an issue has been raised, which would be extremely helpful as the number of GitHub repos goes up, it would be good to have some assurance that what we need the tool do actually works.

Good point! I did run it manually and checked the result, so under this configuration, it works. However automating it would indeed be tricky. A malformed Jira URL would not raise an error on our side, meaning `curl` fires a POST and Jira accepts it, if the webhook is not correct (we cannot authenticate), there will be an error logged that a Jira admin can see, if the webhook is correct and the Automation is misconfigured, we would see it in the Audit log of the Automation action. Still not much if anything that we can do on our side. The Jira URL is also behind a CloudFront that happily accepts the request and returns an `HTTP/2 200`. I don't have a suggestion on how to tackle this.

jdkandersson:
> > Thanks for adding this. Here comes the question, any tests that exercise this? How do we know it works? In this case it is a bit more grey since we wouldn't run this on CI/CD. Although, if we want to use it to rely on getting notifications that an issue has been raised, which would be extremely helpful as the number of GitHub repos goes up, it would be good to have some assurance that what we need the tool do actually works.
> 
> Good point! I did run it manually and checked the result, so under this configuration, it works. However automating it would indeed be tricky. A malformed Jira URL would not raise an error on our side, meaning `curl` fires a POST and Jira accepts it, if the webhook is not correct (we cannot authenticate), there will be an error logged that a Jira admin can see, if the webhook is correct and the Automation is misconfigured, we would see it in the Audit log of the Automation action. Still not much if anything that we can do on our side. The Jira URL is also behind a CloudFront that happily accepts the request and returns an `HTTP/2 200`. I don't have a suggestion on how to tackle this.

I can think of a way, just not simple so it might not be worth doing it. We could have something deployed that has access to the Jira board that can be triggered similar to how we trigger Jira, this would be triggered after the Jira webhook was triggered, would send an API call to Jira to check that the item was created and then return a response to the original webhook to confirm or deny it worked. This would involve something running with network connectivity to Jira we would have to build that can accept webhooks. We have a similar problem for the GitHub runner charm were we don't really have a choice but to implement something like this.

Another thought, how does the action get access to Jira? Couldn't we use the same methodology to send an API call to Jira to query the created item? Surely if create and update works then get would also work?

merkata:
> > > Thanks for adding this. Here comes the question, any tests that exercise this? How do we know it works? In this case it is a bit more grey since we wouldn't run this on CI/CD. Although, if we want to use it to rely on getting notifications that an issue has been raised, which would be extremely helpful as the number of GitHub repos goes up, it would be good to have some assurance that what we need the tool do actually works.
> > 
> > 
> > Good point! I did run it manually and checked the result, so under this configuration, it works. However automating it would indeed be tricky. A malformed Jira URL would not raise an error on our side, meaning `curl` fires a POST and Jira accepts it, if the webhook is not correct (we cannot authenticate), there will be an error logged that a Jira admin can see, if the webhook is correct and the Automation is misconfigured, we would see it in the Audit log of the Automation action. Still not much if anything that we can do on our side. The Jira URL is also behind a CloudFront that happily accepts the request and returns an `HTTP/2 200`. I don't have a suggestion on how to tackle this.
> 
> I can think of a way, just not simple so it might not be worth doing it. We could have something deployed that has access to the Jira board that can be triggered similar to how we trigger Jira, this would be triggered after the Jira webhook was triggered, would send an API call to Jira to check that the item was created and then return a response to the original webhook to confirm or deny it worked. This would involve something running with network connectivity to Jira we would have to build that can accept webhooks. We have a similar problem for the GitHub runner charm were we don't really have a choice but to implement something like this.
> 
> Another thought, how does the action get access to Jira? Couldn't we use the same methodology to send an API call to Jira to query the created item? Surely if create and update works then get would also work?

Good points, I'm not sure about the first one, it seems it will require some effort and at that point I don't know whether it's worth it, I'd like to see some results first :)

For the second - the methodology we're using is via the Jira Automation, so you can imagine that the Jira API is "wrapped", you don't call actual API endpoints, the Automation does it for you. The current approach is that on issue create or close, you hit the webhook URL with a payload, then the Action does a JQL query against the issue title and respectively creates/closes a Jira story.

If we really start actively using github issues and use Jira only for the process, there are options - the ones you mentioned, writing something that makes use of direct API calls (they return a JSON body that can be inspected) or go with a Jira plugin that can do all of these things for you.

So far I'd like to explore this, as it is quick and dirty, that comes with positive and negative sides. I've dealt with Jira API in the past and prefer to stay away from it if possible.

gregory-schiano:
> > > > Thanks for adding this. Here comes the question, any tests that exercise this? How do we know it works? In this case it is a bit more grey since we wouldn't run this on CI/CD. Although, if we want to use it to rely on getting notifications that an issue has been raised, which would be extremely helpful as the number of GitHub repos goes up, it would be good to have some assurance that what we need the tool do actually works.
> > > 
> > > 
> > > Good point! I did run it manually and checked the result, so under this configuration, it works. However automating it would indeed be tricky. A malformed Jira URL would not raise an error on our side, meaning `curl` fires a POST and Jira accepts it, if the webhook is not correct (we cannot authenticate), there will be an error logged that a Jira admin can see, if the webhook is correct and the Automation is misconfigured, we would see it in the Audit log of the Automation action. Still not much if anything that we can do on our side. The Jira URL is also behind a CloudFront that happily accepts the request and returns an `HTTP/2 200`. I don't have a suggestion on how to tackle this.
> > 
> > 
> > I can think of a way, just not simple so it might not be worth doing it. We could have something deployed that has access to the Jira board that can be triggered similar to how we trigger Jira, this would be triggered after the Jira webhook was triggered, would send an API call to Jira to check that the item was created and then return a response to the original webhook to confirm or deny it worked. This would involve something running with network connectivity to Jira we would have to build that can accept webhooks. We have a similar problem for the GitHub runner charm were we don't really have a choice but to implement something like this.
> > Another thought, how does the action get access to Jira? Couldn't we use the same methodology to send an API call to Jira to query the created item? Surely if create and update works then get would also work?
> 
> Good points, I'm not sure about the first one, it seems it will require some effort and at that point I don't know whether it's worth it, I'd like to see some results first :)
> 
> For the second - the methodology we're using is via the Jira Automation, so you can imagine that the Jira API is "wrapped", you don't call actual API endpoints, the Automation does it for you. The current approach is that on issue create or close, you hit the webhook URL with a payload, then the Action does a JQL query against the issue title and respectively creates/closes a Jira story.
> 
> If we really start actively using github issues and use Jira only for the process, there are options - the ones you mentioned, writing something that makes use of direct API calls (they return a JSON body that can be inspected) or go with a Jira plugin that can do all of these things for you.
> 
> So far I'd like to explore this, as it is quick and dirty, that comes with positive and negative sides. I've dealt with Jira API in the past and prefer to stay away from it if possible.

Putting my 2 cents here.
What do we want to test ?

- We surely don't want to test that Jira Automation works as intendent, that's not the purpose here, and should be tested by Jira themselves, we need to trust and assume that Jira is working properly.
- We could want to test that authentication is working, therefore we could try to find a simple API call in Jira that return like the account or Jira version, and call this endpoint in a test to check if the token is working properly and is linked to the expected account. This would be quite trivial if there is such endpoint on Jira and the token is usable for this api call
- We want to ensure that the request we make is sending the expected payload. For this we could start a mock webserver (that accepts any call and output the call in a log file), and call this workflow with forged input (inluding the URL that points to the mock webserver). Then we check that the logs print what we expect (meaning, type of query, headers, body and url with query params).

So in my opinion there are 2 tests we can do:
- Validate authentication works and is linked to the expected account
- Validate the call to Jira and the expected payload

.github/workflows/jira.yaml
gregory-schiano: Also thinking twice about it, in my opinion it would be better to "hardcode" the hostname, and eventually the path, and just store the token as secret.
It won't be very reusable but I don't like mixing real secrets with other data

jdkandersson:
The argument to do this as a trial for now is a good one, we can consider more extensive testing once we are confident this is what we want for the team

.github/workflows/jira.yaml
```diff
@@ -0,0 +1,44 @@
...
+          # Determine Bug/Enhancement
+          if ${{ contains(github.event.*.labels.*.name, 'Bug') }}; then
+            type=bug
+          else
+            type=story
+          fi
+
+          # send JIRA request
+
+          # Canonical discourages discussions in JIRA on public project issues.
```
jdkandersson, line 30: Should this be in this file? Looks a copy paste from somewhere?

merkata: It is directly taken from the original, I've kept all descriptions (only fixed formatting and typos).

merkata: The link to the original is also present in the file.

.github/workflows/jira.yaml
```diff
@@ -0,0 +1,44 @@
...
+          # It is NOT recommended to put description in JIRA in order to
+          # push conversation to GitHub. If strongly required,
+          # replace '--arg body ""' with '--arg body "$ISSUE_DESCRIPTION"'
+
+          data=$( jq -n \
+                  --arg title "$ISSUE_TITLE" \
+                  --arg url '${{ github.event.issue.html_url }}' \
+                  --arg submitter '${{ github.event.issue.user.login }}' \
+                  --arg body "$ISSUE_DESCRIPTION" \
```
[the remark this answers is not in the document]
merkata: Yes, I can add an additional description to justify why we should include it. Without adding the issue body, the issue in Jira has only a link to the github issue and the github username of whoever created it. I'd like to have the full information from Github copied over to Jira and also for it to address our standards, having a description, ways to reproduce, acceptance criteria.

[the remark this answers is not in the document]
merkata: Needless to say, discussion on issues will remain public and on Github, I just don't want information to be lost between to two media.

gregory-schiano (approved):
LGTM, as discussed let's try with this action and if we're satisfied we'll cover with some tests

gtrkiller (approved):
LGTM

PR #92, opened 2023-02-05
get requires integration test to run on failure

jdkandersson:
Changes the required status checks to run even if a job fails so that merging is blocked


nrobinaubertin (approved):

merkata (approved):
Very good, thanks for fixing!

PR #93, opened 2023-02-10
Fix PRs running from forks

arturo-seijas:
Fixes the following actions when running from forks:
* Trivy scan
* Docs publishing in draft mode
* Lib checks

Runs:
* https://github.com/canonical/indico-operator/actions/runs/4143038228
* https://github.com/canonical/indico-operator/actions/runs/4143038232

### DIFF
--- a/.github/workflows/build_images.yaml
+++ b/.github/workflows/build_images.yaml
@@ -79,6 +79,7 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
         if: ${{ github.event.pull_request.head.repo.fork }}
+      # Trivy requires and OCI exporte type to run the scan
       - name: Build image as tarball
         uses: docker/build-push-action@v3
         if: ${{ github.event.pull_request.head.repo.fork }}
--- a/.github/workflows/build_images.yaml
+++ b/.github/workflows/build_images.yaml
@@ -65,6 +65,7 @@ jobs:
           file: ${{ inputs.working-directory }}/${{ matrix.image }}.Dockerfile
       - name: Run Github Trivy Image Action
         uses: aquasecurity/trivy-action@master
+        if: ${{ !github.event.pull_request.head.repo.fork }}
         with:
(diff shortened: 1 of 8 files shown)


nrobinaubertin (dismissed):

.github/workflows/build_images.yaml
```diff
@@ -78,25 +79,26 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
         if: ${{ github.event.pull_request.head.repo.fork }}
+      # Trivy requires and OCI exporter type to run the scan
```
nrobinaubertin, line 82: s/and/an

amandahla (dismissed):
Nice!

.github/workflows/build_images.yaml
arturo-seijas: Fixed, thx

nrobinaubertin (dismissed):

nrobinaubertin (dismissed):

weiiwang01 (dismissed):

amandahla (approved):

mthaddon (approved):

jdkandersson:
We should consider this carefully. It is beneficial for the CI-CD to be able to run for forks, although it isn't clear whether that should result in being able to merge into main. If we merge into main from a fork, anything that is disabled may have issues with it not uncovered by the disabled checks. An alternative is to require any PR forks to be merged into another branch on the repo, get it to run all the checks and then merge into main I'll write a note in the contributing guide where we can debate and link it here.

jdkandersson:
It would also have been good to write a test for the operator workflows running in a fork. I'm not sure whether this would work, an initial idea is to use the GitHub cli to create a fork, trigger a PR and then query the status of that PR: https://cli.github.com/manual/gh_repo_fork

Without this test, there might be changes in the future that break the running of the CI/CD in forks again. An alternative to an automated test would be a list of manual things that need to be checked for each PR, like that the workflows run properly on a fork. That might get out of hand quickly

PR #94, opened 2023-02-16
Add support for building rocks

arturo-seijas:
Test run at https://github.com/canonical/indico-operator/actions/runs/4223929907/jobs/7334246348
Regression https://github.com/canonical/indico-operator/actions/runs/4224017068/jobs/7334445485

### DIFF
--- a/.github/workflows/build_rocks.yaml
+++ b/.github/workflows/build_rocks.yaml
@@ -40,7 +40,8 @@ jobs:
         working-directory: ${{ inputs.working-directory }}
         run: |
           echo "ROCK_PATHS=$(find . -type f -name rockcraft.yaml | xargs realpath | xargs dirname)" >> $GITHUB_ENV
-          echo "ROCKS=$(find . -type f -name rockcraft.yaml | xargs -l yq '.name' | jq -Rsc '. / "\n" - [""]')" >> $GITHUB_ENV
+          ROCKS=$(find . -type f -name rockcraft.yaml | xargs -l yq '.name' | jq -Rsc '. / "\n" - [""]')
+          echo "ROCKS=$ROCKS" >> $GITHUB_ENV
   build-rocks:
     name: Build rock
     runs-on: ${{ inputs.runs-on }}
--- a/.github/workflows/build_rocks.yaml
+++ b/.github/workflows/build_rocks.yaml
@@ -86,7 +86,7 @@ jobs:
         if: ${{ github.event.pull_request.head.repo.fork }}
         run: |
           skopeo copy oci-archive:$(ls ${{ matrix.path }}/${{ matrix.image }}*.rock) dir:/tmp/$IMAGE_NAME.tar
-        echo "IMAGE_TAR=${{ matrix.image }}.tar" >> $GITHUB_ENV
+          echo "IMAGE_TAR=${{ matrix.image }}.tar" >> $GITHUB_ENV
       - name: Run Github Trivy Image Action
         uses: aquasecurity/trivy-action@master
         with:
--- a/.github/workflows/build_rocks.yaml
+++ b/.github/workflows/build_rocks.yaml
@@ -39,7 +39,7 @@ jobs:
       - name: Get rock paths
         working-directory: ${{ inputs.working-directory }}
         run: |
-          echo "ROCK_PATHS=$(find . -type f -name rockcraft.yaml | xargs realpath | xargs dirname" >> $GITHUB_ENV
+          echo "ROCK_PATHS=$(find . -type f -name rockcraft.yaml | xargs realpath | xargs dirname)" >> $GITHUB_ENV
           echo "ROCKS=$(find . -type f -name rockcraft.yaml | xargs -l yq '.name' | jq -Rsc '. / "\n" - [""]')" >> $GITHUB_ENV
   build-rocks:
     name: Build rock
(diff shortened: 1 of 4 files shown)


weiiwang01 (dismissed):

.github/workflows/build_rocks.yaml
```diff
@@ -0,0 +1,107 @@
...
+      - name: Build rock
+        uses: canonical/craft-actions/rockcraft-pack@main
+        with:
+          path: ${{ matrix.path }}
+      - name: Upload ROCK to ghcr.io
+        if: ${{ !github.event.pull_request.head.repo.fork }}
+        run: |
+          IMAGE_NAME=$(yq '.name' ${{ matrix.path }}/rockcraft.yaml)
+          IMAGE_REF=${{ inputs.registry }}/${{ inputs.owner }}/$IMAGE_NAME:${{ github.run_id }}
+          skopeo --insecure-policy copy oci-archive:$(ls ${{ matrix.path }}/${{ matrix.image }}*.rock) docker://$IMAGE_REF --dest-creds "${{ github.actor }}:${{ secrets.GITHUB_TOKEN }}"
```
gregory-schiano, line 69: No risks of catching multiple files with `${{ matrix.path }}/${{ matrix.image }}*.rock` ?

.github/workflows/build_rocks.yaml
```diff
@@ -0,0 +1,107 @@
...
+          severity: 'CRITICAL,HIGH'
+        env:
+          TRIVY_USERNAME: ${{ github.actor }}
+          TRIVY_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
+      # GitHub doesn't currently support pushing images in the Docker registry when opening a PR from a fork
+      # so an artifact is published instead
+      - name: Build rock as tarball
+        if: ${{ github.event.pull_request.head.repo.fork }}
+        run: |
+          skopeo copy oci-archive:$(ls ${{ matrix.path }}/${{ matrix.image }}*.rock) docker-archive::/tmp/$IMAGE_NAME.tar:$IMAGE_NAME:${{ github.run_id }}
```
gregory-schiano, line 88: Same as above

gregory-schiano: Is the double ":" is required ?

.github/workflows/build_rocks.yaml
```diff
@@ -0,0 +1,107 @@
...
+        run: |
+          skopeo copy oci-archive:$(ls ${{ matrix.path }}/${{ matrix.image }}*.rock) dir:/tmp/$IMAGE_NAME.tar
+      - name: Run Github Trivy Image Action
+        if: ${{ github.event.pull_request.head.repo.fork }}
+        uses: aquasecurity/trivy-action@master
+        with:
+          input: ${{ matrix.image }}.tar
+          trivy-config: ${{ inputs.trivy-image-config }}
+          exit-code: '1'
+          severity: 'CRITICAL,HIGH'
```
gregory-schiano, line 107: Isn't some factoring possible to avoid having this action twice in this workflow ?

.github/workflows/integration_test.yaml
```diff
@@ -157,10 +157,36 @@ jobs:
...
+    outputs:
+      images: ${{ env.IMAGES }}
+    steps:
+      - name: Get rocks or fallback to Docker images
+        run: |
+          if [ ${{ needs.build-rocks.outputs.images != '[]' }} ]; then
+            IMAGES='${{ needs.build-rocks.outputs.images }}'
+          else
+            IMAGES='${{ needs.build-images.outputs.images }}'
+          fi
```
gregory-schiano, line 183: This means we don't support "mixed" situation where we have dockerfiles and rockcraft files ?
Isn't it more flexible to "merge" both output ?

.github/workflows/build_rocks.yaml
arturo-seijas: No, at the moment only one descriptor named rockcraft.yaml is allowed so we'll only be getting one *.rock per directory https://github.com/canonical/rockcraft/issues/185

gregory-schiano: I'm wondering if you won't have multiple .rock file if there are multiple platforms supported in the rockcraft.yaml

.github/workflows/build_rocks.yaml
arturo-seijas: Perhaps. I'll try

.github/workflows/integration_test.yaml
arturo-seijas: We don't. It is doable but it would be much more complex and we'd need to handle clash scenarios too. I'm not sure it's worth it considering most of the charms will only use one image and rocks will become the default at some point.

gregory-schiano: Well until`rockcraft` isn't mature, we could have a scenario where only a part of the images are available through rocks which requires a mix of rocks and dockerfile
But that's an hypothetical case, and in such case we could decide to keep dockerfile only until the missing support until `rockcraft` supports it.

arturo-seijas: I'd add that support only if we need it, but oc this is my personal opinion

.github/workflows/build_rocks.yaml
arturo-seijas: Not for out use cases. The same would be applicable to the file resulting from packing the charm

gregory-schiano: Yes I agree, I don't want to address this case, but if it happens this workflow will fail because the $(ls...) will return multiple files
So I would at list restrict to the first result of the `ls` to make sure the workflow doesn't just fail

arturo-seijas: It would be easier to catch an issue if it fails when compared to using an image not expected by the user, IMHO

gregory-schiano: Ok we'll see how to manage that if someday it breaks due to multiple platform

gregory-schiano (approved):
LGTM

weiiwang01 (approved):

PR #98, opened 2023-03-01
feat: allow skipping of publish docs

yanksyoon:
Since the automation for uploading documentation is on hold, this PR introduces `publish-docs` input variable to the workflow that defaults to false to skip Charmhub documentation automation workflow. No changes have to be made to repositories using the workflow.

Link to working [skipped workflow](https://github.com/canonical/wordpress-k8s-operator/actions/runs/4302974557/jobs/7502135654).

### DIFF
--- a/.github/workflows/publish_charm.yaml
+++ b/.github/workflows/publish_charm.yaml
@@ -10,7 +10,9 @@ on:
       images:
         type: string
         description: 'Docker images'
-
+      publish-docs:
+        type: boolean
+        description: 'Publish charm documentations to Charmhub'
 env:
   REGISTRY: ghcr.io
   OWNER: ${{ github.repository_owner }}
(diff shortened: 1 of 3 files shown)


arturo-seijas (commented):
Under which scenario is skipping the step needed? Right now it runs as draft so no docs are published

yanksyoon:
> Under which scenario is skipping the step needed? Right now it runs as draft so no docs are published

It runs as draft but for wordpress-k8s charm there's currently no need to run it in dry_run mode either but it will still fail due to missing Charmhub credentials.
For such charms I think it should be okay to not have credentials set up in the repository whilst letting the documentations still be merged until we have a pass on automating documentations :)

jdkandersson:
> > Under which scenario is skipping the step needed? Right now it runs as draft so no docs are published
> 
> It runs as draft but for wordpress-k8s charm there's currently no need to run it in dry_run mode either but it will still fail due to missing Charmhub credentials. For such charms I think it should be okay to not have credentials set up in the repository whilst letting the documentations still be merged until we have a pass on automating documentations :)

What is the issue with setting up credentials? Adding this feature increases the complexity of our CI/CD. In the future, I can only see either that the action is removed entirely or used in all repositories depending on where we land with the spec being reviewed

yanksyoon:
Makes sense, thanks for the explanation, closing.

PR #99, opened 2023-03-02
add missing brackets

jdkandersson:
A few of the `always()` et al checks were missing brackets which meant that they weren't working. This PR fixes those

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -230,7 +230,7 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - integration-test
-    if: always()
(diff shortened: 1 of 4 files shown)


weiiwang01:
I don't think brackets are necessary in the `if` statement
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif
> When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) because GitHub automatically evaluates the if conditional as an expression. For more information, see "[Expressions](https://docs.github.com/en/actions/learn-github-actions/expressions)."

jdkandersson:
As noted by Weii, this isn't needed

PR #100, opened 2023-03-02
ensure past runs are cancelled

jdkandersson:
The GitHub organisation only has at most 60 concurrent runners. This change ensures that any outstanding actions from a previous run are cancelled. Without those cancellations, the 60 runners can quickly become oversubscribed

This change also limits the concurrency of the tests of `operator-workflow`.

I have deliberately pushed a new change before a previous run was finished to check that the cancellation works. Creating an automated test for that would be cumbersome so the proposal is not to do that in this case.


weiiwang01 (dismissed):

mthaddon (dismissed):

nrobinaubertin (dismissed):

jdkandersson:
Putting this back to draft, there are some weird things going on with GitHub actions

jdkandersson:
I think what might be happening is that there is a past run that is pending due to some problems on GitHub at the time which can't be cancelled (an error occurs on attempting to cancel it) which is resulting in the actions not getting executed becauyse they have to cancel that run before being able to start a new one

.github/workflows/integration_test.yaml
```diff
@@ -141,6 +141,10 @@ env:
   REGISTRY: ghcr.io
   OWNER: ${{ github.repository_owner }}
 
+concurrency:
+  group: ${{ github.workflow }}-integration-tests-${{ github.ref }}
```
weiiwang01, line 145: I think maybe we should add a `operator-workflows` prefix to the concurrency group name to prevent potential collisions with other workflows.

jdkandersson: Good point, added, thanks

weiiwang01 (approved):

jdkandersson:
Feel free to merge this if it gets another approval and no comment

amandahla (approved):
LGTM

PR #103, opened 2023-03-09
Add an integration test for rockcraft based OCI image build

weiiwang01:


tests/workflows/integration/test-rock/rockcraft.yaml
```diff
@@ -0,0 +1,16 @@
+name: test
+summary: Test rock.
+description: Test rock.
+version: "1.0"
+base: ubuntu:22.04
```
arturo-seijas, line 5: Perhaps using "bare" will be faster

arturo-seijas (commented):
Just a minor suggestion

arturo-seijas (dismissed):
LGTM

amandahla (dismissed):
LGTM

arturo-seijas (dismissed):
LGTM

arturo-seijas (dismissed):
LGTM

arturo-seijas (approved):
LGTM

merkata:
Just to be on the safe side - the logic is that in the integration workflow we skip building the rock and it is built in the integration-rock workflow, correct? Looks good!

merkata (approved):

weiiwang01:
> Just to be on the safe side - the logic is that in the integration workflow we skip building the rock and it is built in the integration-rock workflow, correct? Looks good!

Yes. In the integration, the OCI image is built by docker.

PR #108, opened 2023-03-19
Fix the rockcraft workflow in PR from forked repositories

weiiwang01:


.github/workflows/build_rocks.yaml
```diff
@@ -61,32 +61,29 @@ jobs:
...
-        if: ${{ !github.event.pull_request.head.repo.fork }}
+      - name: Extract rock information
         run: |
           IMAGE_NAME=$(yq '.name' ${{ matrix.path }}/rockcraft.yaml)
           IMAGE_REF=${{ inputs.registry }}/${{ inputs.owner }}/$IMAGE_NAME:${{ github.run_id }}
-          skopeo --insecure-policy copy oci-archive:$(ls ${{ matrix.path }}/${{ matrix.image }}*.rock) docker://$IMAGE_REF --dest-creds "${{ github.actor }}:${{ secrets.GITHUB_TOKEN }}"
           echo "IMAGE_NAME=$IMAGE_NAME" >> $GITHUB_ENV
           echo "IMAGE_REF=$IMAGE_REF" >> $GITHUB_ENV
+      - name: Upload rock to ghcr.io
+        if: ${{ !github.event.pull_request.head.repo.fork }}
```
jdkandersson, line 71: Are there any problems with not uploading the rock if the run is from a fork? What would we use that for?

.github/workflows/build_rocks.yaml
```diff
@@ -61,32 +61,29 @@ jobs:
...
+        run: |
+          skopeo --insecure-policy copy oci-archive:$(ls ${{ matrix.path }}/*.rock) docker://$IMAGE_REF --dest-creds "${{ github.actor }}:${{ secrets.GITHUB_TOKEN }}"
```
jdkandersson, line 73: I realise this was already like this, do you know the reason for requiring this argument? `--insecure-policy`?

.github/workflows/workflow_test.yaml
```diff
@@ -62,7 +63,6 @@ jobs:
       - dockerfile
       - integration
       - integration-rock
-      - publish
```
jdkandersson, line 65: What is the reason for this removal please?

.github/workflows/build_rocks.yaml
weiiwang01: When running from a fork, the Github action does not have sufficient permission to push container images to the ghcr registry.

.github/workflows/build_rocks.yaml
weiiwang01: Just quote from the skopeo document:

> --insecure-policy
> 
> Adopt an insecure, permissive policy that allows anything. This obviates the need for a policy file.

I think this is just to simplify the push process.

.github/workflows/workflow_test.yaml
weiiwang01: When running in a PR from a forked repository, the `publish` job will be skipped. That's why it's removed here.

.github/workflows/build_rocks.yaml
jdkandersson: And what are the consequences of not having the image in the ghcr registry? Is it used later on in the workflow?

.github/workflows/workflow_test.yaml
jdkandersson: Why is it skipped? Is it because the workflow doesn't have access to the required secrets? We have an open PR that discusses this topic and the best we have come up with so far is that PRs from forks should be merged into a separate branch after initial review so that the full test suit can be run: https://github.com/canonical/is-charms-contributing-guide/pull/52

jdkandersson: If we don't execute all the checks then any PR from a fork could introduce breaking changes to the parts of `operator-workflows` that are not tested which results in our workflows breaking in repositories that rely on it working

weiiwang01: Actually, I am not a fan of creating a new branch and PR ourself every time an outside contributor wants to contribute.

I think another approach for this problem is to use special comments to trigger actions for PR initiated by outside contributors. You can find an example here: https://github.com/orgs/community/discussions/25389

But implementing that can not be done very quickly, for now, we can just skip the check since this PR https://github.com/canonical/operator-workflows/pull/107 is waiting on use.

.github/workflows/build_rocks.yaml
```diff
@@ -61,32 +61,29 @@ jobs:
...
       # GitHub doesn't currently support pushing images in the Docker registry when opening a PR from a fork
       # so an artifact is published instead
       - name: Build rock as tarball
         if: ${{ github.event.pull_request.head.repo.fork }}
         run: |
-          skopeo copy oci-archive:$(ls ${{ matrix.path }}/${{ matrix.image }}*.rock) docker-archive:/tmp/$IMAGE_NAME.tar:$IMAGE_NAME:${{ github.run_id }}
+          skopeo copy oci-archive:$(ls ${{ matrix.path }}/*.rock) docker-archive:./$IMAGE_NAME.tar:$IMAGE_NAME:${{ github.run_id }}
+          echo "IMAGE_TAR=/github/workspace/$IMAGE_NAME.tar" >> $GITHUB_ENV
```
arturo-seijas, line 80: Is this variable actually used?

.github/workflows/build_rocks.yaml
gregory-schiano: I think in such case the image is using a locally build image, which is then pushed to a local registry in microk8s therefore ghcr repository is not needed.
The final image will anyway be built and pushed to ghcr when the merge on main will occur

.github/workflows/build_rocks.yaml
```diff
@@ -61,32 +61,29 @@ jobs:
...
+        if: ${{ !github.event.pull_request.head.repo.fork }}
+        run: |
+          skopeo --insecure-policy copy oci-archive:$(ls ${{ matrix.path }}/*.rock) docker://$IMAGE_REF --dest-creds "${{ github.actor }}:${{ secrets.GITHUB_TOKEN }}"
       # GitHub doesn't currently support pushing images in the Docker registry when opening a PR from a fork
       # so an artifact is published instead
       - name: Build rock as tarball
         if: ${{ github.event.pull_request.head.repo.fork }}
         run: |
-          skopeo copy oci-archive:$(ls ${{ matrix.path }}/${{ matrix.image }}*.rock) docker-archive:/tmp/$IMAGE_NAME.tar:$IMAGE_NAME:${{ github.run_id }}
+          skopeo copy oci-archive:$(ls ${{ matrix.path }}/*.rock) docker-archive:./$IMAGE_NAME.tar:$IMAGE_NAME:${{ github.run_id }}
```
gregory-schiano, line 79: I don't know if github actions backquote the variables but if `matrix.path` contains a space there'll be an issue with the `ls` command.
Better protect it with double-quotes
`$(ls "${{ matrix.path }}/*.rock")`

.github/workflows/build_rocks.yaml
weiiwang01: In a PR from a forked repository, the image is uploaded as an artifact instead of push to ghcr registry. And in the following integration tests, the image is downloaded from artifacts or pull from ghcr registry according to the initiator of the PR.

.github/workflows/build_rocks.yaml
weiiwang01: Yes, it was used in the final step `Run Github Trivy Image Action`.

.github/workflows/build_rocks.yaml
weiiwang01: Yes, it could case a problem when the `matrix.path`. But it will need some small adjustment since put the wildcard in quotes will stop the wildcard expanding.
I think this will work: `$(ls "${{ matrix.path }}"/*.rock)`

weiiwang01: Updated, thanks!

arturo-seijas (dismissed):
LGTM

.github/workflows/build_rocks.yaml
weiiwang01: If the image was pushed to the microk8s registry, it will be gone once the `build_rocks` job finished.

.github/workflows/workflow_test.yaml
jdkandersson: As discussed, let's change the check to allow for skipped in the case of forks and we'll add a story to the backlog with the comment to trigger a workflow run

weiiwang01: I just add the `publish` workflow back to the check dependencies.

.github/workflows/workflow_test.yaml
```diff
@@ -75,4 +76,4 @@ jobs:
           [ '${{ needs.dockerfile.outputs.docker-lint-outcome }}' = 'success' ] || (echo docker-lint failed && false)
           [ '${{ needs.integration.result }}' = 'success' ] || (echo integration failed && false)
           [ '${{ needs.integration-rock.result }}' = 'success' ] || (echo integration-rock failed && false)
-          [ '${{ needs.publish.result }}' = 'success' ] || (echo publish failed && false)
+          [ '${{ needs.publish.result }}' != 'failure' ] || (echo integration-rock failed && false)
```
jdkandersson, line 79: Should this be `publish` rather than `integration-rock`?

weiiwang01: Yes, indeed. Updated, thanks!

jdkandersson (approved):

arturo-seijas (approved):
LGTM

PR #109, opened 2023-03-22
[DO NOT MERGE] Isd 420 investigate GitHub workflow caching

merkata:
This PR relates to the spike on Github caching and how it can speed up our workflows. It is a POC and design should be discussed if we decide to move forward with using caching for some components in our CI/CD. Currently there is caching for unit tests and integration tests, the integration tests include caching of dependencies and built rocks and a built charm. In order for the integration tests to make use of the built charm, there needs to be a local change in the integration tests for the charm repo under test. The speed improvements vary from a minute on unit tests to several minutes (built rocks 5-8 mins and built charm 4-5 mins). The cache is invalidated either weekly or on a dependency file change. This PR is for review and feedback only.

### DIFF
--- a/.github/workflows/build_rocks.yaml
+++ b/.github/workflows/build_rocks.yaml
@@ -63,9 +63,11 @@ jobs:
           echo "date=$(/bin/date -u "+%Y%W")" >> $GITHUB_OUTPUT
       - uses: actions/cache@v3
         id: cache
+        env:
+          ROCK_PATH: ${{ matrix.path }}
         with:
           path: ${{ matrix.path }}
-          key: rock-cache-${{ steps.get-date.outputs.date }}-${{ hashFiles(${{ matrix.path }}/rockcraft.yaml) }}
+          key: rock-cache-${{ steps.get-date.outputs.date }}-${{ hashFiles($ROCK_PATH/rockcraft.yaml) }}
       - name: Build rock
         if: steps.cache.outputs.cache-hit != 'true'
         uses: canonical/craft-actions/rockcraft-pack@main
--- a/.github/workflows/build_rocks.yaml
+++ b/.github/workflows/build_rocks.yaml
@@ -65,7 +65,7 @@ jobs:
         id: cache
         with:
           path: ${{ matrix.path }}
-        key: rock-cache-${{ steps.get-date.outputs.date }}-${{ hashFiles(${{ matrix.path }}/rockcraft.yaml) }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build_rocks.yaml
```diff
@@ -57,7 +57,17 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
+      - name: Get Date
+        id: get-date
+        run: |
+          echo "date=$(/bin/date -u "+%Y%W")" >> $GITHUB_OUTPUT
+      - uses: actions/cache@v3
+        id: cache
+        with:
+          path: ${{ matrix.path }}
```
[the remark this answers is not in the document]
merkata: I haven't looked into that and it seems rockcraft specific - the Github cache is basically a list of files and directories that get tar-ed on cache creation and untar-ed when accessing the cache, so how you interact with the files in the cache is up to the tool, like rockcraft in that case.

.github/workflows/integration_test_run.yaml
```diff
@@ -204,6 +208,17 @@ jobs:
       - name: Pre-run script
         if: ${{ inputs.pre-run-script != '' }}
         run: bash ${{ inputs.pre-run-script }}
+      - uses: actions/cache@v3
+        id: cache-integration
+        with:
+          path: |
+            '**/*.charm'
+            .tox
+          key: integration-cache-${{ steps.get-date.outputs.date }}-${{ hashFiles('**/charmcraft.yaml', '**/metadata.yaml', '**/actions.yaml', '**/config.yaml', '**/src') }}
```
arturo-seijas, line 217: Perhaps we should include the integration test dir too? It is my understanding that this charm is used to run those.

.github/workflows/build_rocks.yaml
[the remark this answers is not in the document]
arturo-seijas: Perhaps it's worth a feature request. It would be a more performant solution

.github/workflows/integration_test_run.yaml
jdkandersson, line 217: Any particular reason for including the date? Is it so that dependencies are re-installed frequently, eg., if there is a new patch version available?

merkata: The date is to ensure the cache gets invalidated at some point, in the particular case it lasts at most one week. The reason behind it is that not all dependencies we have are pinned to specific versions and we don't bump them in the files that are hashed for the cache key, so there needs to be a safety mechanism to invalidate the cache at some point, otherwise it will get stale after a while.

PR #112, opened 2023-04-06
GitHub token

jdkandersson:
This fixes a problem where `upload-charm-docs` got the Charmhub token instead of the GitHub token

I thought about adding tests for this, it wouldn't be straight forward, I'll add it to our backlog. Since our CI is blocked by this not working I propose to merge this without adding an integration test initially

### DIFF
--- a/.github/workflows/promote_charm.yaml
+++ b/.github/workflows/promote_charm.yaml
@@ -51,6 +51,7 @@ jobs:
           discourse_api_key: ${{ secrets.DISCOURSE_API_KEY }}
           dry_run: true
           github_token: ${{ secrets.GITHUB_TOKEN }}
+          base_branch: main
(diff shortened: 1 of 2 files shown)


jdkandersson:
Ok, I tested this for Wordpress and it works as intended. It actually picked up some conflicting changes which is interesting: https://github.com/canonical/wordpress-k8s-operator/pull/66

yhaliaw (approved):

merkata (approved):
Looks good, I assume that was a typo in the first place, passing in CHARMHUB_TOKEN to a github_token key?

jdkandersson:
> Looks good, I assume that was a typo in the first place, passing in CHARMHUB_TOKEN to a github_token key?

yes thats right

PR #115, opened 2023-04-11
Fail the integration test if pre-run script failed

weiiwang01:
Right now, if the `pre-run-script` in the integration test workflow failed, the integration test will proceed without an error.


arturo-seijas (approved):
LGTM

merkata (approved):

PR #125, opened 2023-04-24
Do not require license header for config files

arturo-seijas:
Do not require license header for .woke.yaml and zap config file


nrobinaubertin (approved):

gtrkiller (approved):
LGTM

PR #136, opened 2023-05-12
Add run on self-hosted runner as option for test.yaml

yhaliaw:
Add input called `self-hosted-runner` for test.yaml, when set to true the workflow will be run on self-hosted runners.

### DIFF
--- a/.github/workflows/auto_update_charm_libs.yaml
+++ b/.github/workflows/auto_update_charm_libs.yaml
@@ -32,8 +32,8 @@ jobs:
           branch-name: "chore/auto-libs"
           title: Update charm libraries
           body: |
-            Automated action to fetch latest version of charm libraries. The branch of this PR 
-            will be wiped during the next check. Unless you really know what you're doing, you 
+            Automated action to fetch latest version of charm libraries. The branch of this PR
+            will be wiped during the next check. Unless you really know what you're doing, you
             most likely don't want to push any commits to this branch.
           upsert: true
           ignore-no-changes: true
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -23,7 +23,7 @@ on:
 jobs:
   inclusive-naming-check:
     name: Inclusive naming
-    runs-on: [self-hosted, large]
+    runs-on: ${{ inputs.self-hosted-runner && fromJson('[''self-hosted'', ''large'']') || 'ubuntu-22.04' }}
     defaults:
       run:
         working-directory: ${{ inputs.working-directory }}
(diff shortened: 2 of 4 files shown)


.github/workflows/test.yaml
```diff
@@ -9,7 +9,11 @@ on:
       working-directory:
         type: string
         description: The working directory for jobs
-        default: './'
+        default: "./"
```
merkata, line 12: I'm curious about the change from single to double quotes, why is that?

yhaliaw: The file uses both single quotes and double quotes. I used a formatter (YAML extension by Red Hat on vscode) to enforce using double quotes. Double quotes is chosen as less change was required compared to using single qoutes.

.github/workflows/test.yaml
```diff
@@ -33,7 +37,7 @@ concurrency:
 jobs:
   inclusive-naming-check:
     name: Inclusive naming
-    runs-on: ubuntu-22.04
+    runs-on: ${{ inputs.self-hosted-runner && '[self-hosted, xlarge]' || 'ubuntu-22.04' }}
```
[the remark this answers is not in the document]
yhaliaw: The quotes within GitHub Action expression has to be single quotes according to the documentation: https://docs.github.com/en/actions/learn-github-actions/expressions#literals

.github/workflows/auto_update_charm_libs.yaml
```diff
@@ -32,8 +32,8 @@ jobs:
           branch-name: "chore/auto-libs"
           title: Update charm libraries
           body: |
-            Automated action to fetch latest version of charm libraries. The branch of this PR 
-            will be wiped during the next check. Unless you really know what you're doing, you 
+            Automated action to fetch latest version of charm libraries. The branch of this PR
+            will be wiped during the next check. Unless you really know what you're doing, you
```
arturo-seijas, line 36: I loved this bit "Unless you really know what you're doing"

arturo-seijas (dismissed):
LGTM

.github/workflows/test.yaml
```diff
@@ -9,16 +9,19 @@ on:
       working-directory:
         type: string
         description: The working directory for jobs
-        default: './'
+        default: "./"
+      self-hosted-runner:
+        type: boolean
+        description: Whether to use self-hosted runners to run the jobs.
+        default: false
```
jdkandersson, line 16: Should we make the default to use the self-hosted runners? Or perhaps we do that after we scale them?

jdkandersson (dismissed):

.github/workflows/test.yaml
yhaliaw: I think it might be good to set the default to true. If developers are encountering issues when they are working a PR, they can set it `self-hosted-runner` to false.

jdkandersson (approved):

weiiwang01 (approved):

PR #137, opened 2023-05-12
Validate channels when promoting charms

arturo-seijas:
Validate origin and destinations tracks match when running the promote charm workflow to prevent human error

### DIFF
--- a/.github/workflows/promote_charm.yaml
+++ b/.github/workflows/promote_charm.yaml
@@ -21,10 +21,44 @@ jobs:
   get-runner-image:
     name: Get runner image
     uses: ./.github/workflows/get_runner_image.yaml
+  validate-channels:
+    name: Validate channels
+    needs: get-runner-image
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - run: |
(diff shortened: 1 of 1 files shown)


gregory-schiano:
If I get it well it ensures that the origin and destination tracks stay the same, for example:
Promoting from `latest/edge` to `latest/stable` is allowed but `1.4/edge` can't be promoted to `latest/stable` right ?
How will we manage when we will have a new latest and we want to move the old latest to X.Y track (X.Y being the major and minor version of the previous app version) ?

arturo-seijas:
> If I get it well it ensures that the origin and destination tracks stay the same, for example: Promoting from `latest/edge` to `latest/stable` is allowed but `1.4/edge` can't be promoted to `latest/stable` right ? How will we manage when we will have a new latest and we want to move the old latest to X.Y track (X.Y being the major and minor version of the previous app version) ?

The idea is to stop publishing to tracks without version number is this is relevant. We'll go with `latest` when irrelevant, and with a version number otherwise.

weiiwang01 (approved):

gregory-schiano (approved):
LGTM

PR #138, opened 2023-05-12
Add support for overriding the publish channel

arturo-seijas:
Add support for overriding the publish channel on the test and publish workflow

### DIFF
--- a/.github/workflows/test_and_publish_charm.yaml
+++ b/.github/workflows/test_and_publish_charm.yaml
@@ -3,13 +3,6 @@
 
 name: Test and publish charm
 
-# On push to a "special" branch, we:
-# * always publish to charmhub at latest/edge/branchname
-# * always run tests
-# where a "special" branch is one of main/master or track/**, as
-# by convention these branches are the source for a corresponding
-# charmhub edge channel.
-
 on:
   workflow_call:
     inputs:
--- a/.github/workflows/test_and_publish_charm.yaml
+++ b/.github/workflows/test_and_publish_charm.yaml
@@ -13,6 +13,10 @@ name: Test and publish charm
 on:
   workflow_call:
     inputs:
+      channel:
+        description: Destination channel to push the charm to
+        type: string
+        default: 'latest/edge'
       integration-test-extra-arguments:
         description: Additional arguments to pass to the integration tests
         type: string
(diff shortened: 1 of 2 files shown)


gregory-schiano:
Why do we do that instead of applying the charming-action enforced convention for branch name ?
It feels like we want to do things our own way, which I'm not sure is a good idea... we should unless really good reason use the charming-actions

arturo-seijas:
> Why do we do that instead of applying the charming-action enforced convention for branch name ? It feels like we want to do things our own way, which I'm not sure is a good idea... we should unless really good reason use the charming-actions

That step retrieves `latest/edge` for main and `[track]/edge` for branches matching with `track/*`. Keeping it while allowing overrides will increase complexity and we aren't actually leveraging it

weiiwang01 (approved):

gregory-schiano:
> > Why do we do that instead of applying the charming-action enforced convention for branch name ? It feels like we want to do things our own way, which I'm not sure is a good idea... we should unless really good reason use the charming-actions
> 
> That step retrieves `latest/edge` for main and `[track]/edge` for branches matching with `track/*`. Keeping it while allowing overrides will increase complexity and we aren't actually leveraging it

Why do we need override then ? I'm worried we would get questions on why we don't use the charming actions and chose our "own way".
I'd rather review the way we work to match other charm engineering teams that going our own way. If we have feedback/constrains that can't match the way other teams are doing, let's discuss on the charming action repo if they can allow "override" for example.

arturo-seijas:
> > > Why do we do that instead of applying the charming-action enforced convention for branch name ? It feels like we want to do things our own way, which I'm not sure is a good idea... we should unless really good reason use the charming-actions
> > 
> > 
> > That step retrieves `latest/edge` for main and `[track]/edge` for branches matching with `track/*`. Keeping it while allowing overrides will increase complexity and we aren't actually leveraging it
> 
> Why do we need override then ? I'm worried we would get questions on why we don't use the charming actions and chose our "own way". I'd rather review the way we work to match other charm engineering teams that going our own way. If we have feedback/constrains that can't match the way other teams are doing, let's discuss on the charming action repo if they can allow "override" for example.

[Here's the current behaviour](https://github.com/canonical/charming-actions/tree/main/channel#branch-selection). In order to fit our use case we'd need to enable overriding the track, which would be the same as not using the action at all.

yanksyoon (approved):
LGTM!

merkata:
Sorry I'm a bit late for the discussion, but what is our use case?

arturo-seijas:
> Sorry I'm a bit late for the discussion, but what is our use case?

This is the intended behaviour https://github.com/canonical/indico-operator/pull/267. Will be extracting the indico version and publishing to [indico_version]/edge on merge

PR #139, opened 2023-05-15
Ignore no changes in the auto update charm lib workflow

weiiwang01:


.github/workflows/auto_update_charm_libs.yaml
```diff
@@ -36,3 +36,4 @@ jobs:
             will be wiped during the next check. Unless you really know what you're doing, you 
             most likely don't want to push any commits to this branch.
           upsert: true
+          ignore-no-changes: true
```
jdkandersson, line 39: Any better way to phrase this? What is the option intended for? Perhaps `skip-empty-pr`?

weiiwang01: When calling the `create-pull-request` and there isn't any local changes, the `create-pull-request` action will fail. This option suppress the error when there's no changes in local.

jdkandersson: Do you think we can come up with a clearer name for the option?

weiiwang01: The reason I want to avoid using the word empty here is that empty has a special meaning in git that is the commit exist but there's no changes in the commit. In our situation, the commit doesn't exist which is a little bit different. That why I choose this name.

jdkandersson (approved):

.github/workflows/auto_update_charm_libs.yaml
weiiwang01: I will add more documents and error messages related to this option. Personally speaking, this option shouldn't be used by average users, since user should judge if the pull request is needed and invoke the `create-pull-request` action accordingly.

arturo-seijas (approved):

PR #140, opened 2023-05-26
feat: enable tmate debugging

yanksyoon:
This PR enables the use of tmate debugging on integration test failure if `tmate-debug` has been enabled.

The motivation behind this PR is to reduce the number of branches called `debug-*` and also some problems caused when the user can get locked out of tmate and need to wait for timeout (6 hours) accidentally.

[Example of failing integration test w/ tmate timeout 5 min](https://github.com/canonical/jenkins-k8s-operator/actions/runs/5086686731/jobs/9141417604)

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -103,6 +103,10 @@ on:
         description: Use setup-devstack-swift action to prepare a swift server for testing.
         type: boolean
         default: false
+      tmate-debugging:
+        description: Use tmate debugging session on failure.
+        type: boolean
+        default: false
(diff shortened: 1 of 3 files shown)


jdkandersson (approved):

merkata (approved):
I was about to say that there is one branch with tmate-debugging that is open - it remained open so engineers can change the step which executes `tmate`, I needed it just before the chaos tests for instance. Now seeing that there are several such branches, I think it is best to have it in main. I also like the timeout addition, well done!

gregory-schiano (approved):
LGTM

PR #141, opened 2023-05-26
Publish external images in the publish workflow

weiiwang01:
Current publish workflow only uploads images built by the integration workflow to CharmHub.
This pull request adds a new step in the publish workflow to upload any OCI-image resources with `upstream-source` to CharmHub before publish the new charm version.

### DIFF
--- a/.github/workflows/build_rocks.yaml
+++ b/.github/workflows/build_rocks.yaml
@@ -79,7 +79,7 @@ jobs:
       - name: Build rock as tarball
         if: ${{ github.event.pull_request.head.repo.fork }}
         run: |
-          skopeo copy oci-archive:$(ls "${{ matrix.path }}"/*.rock) docker-archive:./$IMAGE_NAME.tar:localhost:32000/$IMAGE_NAME:latest
+          skopeo copy oci-archive:$(ls "${{ matrix.path }}"/*.rock) docker-archive:./$IMAGE_NAME.tar:localhost:32000/$IMAGE_NAME:${{ github.run_id }}
           echo "IMAGE_TAR=/github/workspace/$IMAGE_NAME.tar" >> $GITHUB_ENV
       - name: Upload image artifact
         uses: actions/upload-artifact@v3
--- a/.github/workflows/build_rocks.yaml
+++ b/.github/workflows/build_rocks.yaml
@@ -79,7 +79,7 @@ jobs:
       - name: Build rock as tarball
         if: ${{ github.event.pull_request.head.repo.fork }}
         run: |
-          skopeo copy oci-archive:$(ls "${{ matrix.path }}"/*.rock) docker-archive:./$IMAGE_NAME.tar:$IMAGE_NAME:${{ github.run_id }}
+          skopeo copy oci-archive:$(ls "${{ matrix.path }}"/*.rock) docker-archive:./$IMAGE_NAME.tar:localhost:32000/$IMAGE_NAME:latest
           echo "IMAGE_TAR=/github/workspace/$IMAGE_NAME.tar" >> $GITHUB_ENV
       - name: Upload image artifact
         uses: actions/upload-artifact@v3
--- a/.github/workflows/publish_charm.yaml
+++ b/.github/workflows/publish_charm.yaml
@@ -84,7 +84,7 @@ jobs:
           echo "External Images: $IMAGES"
           echo "IMAGES=$IMAGES" >> $GITHUB_ENV
   publish-external-images:
-    name: Push Externam Images to Charmhub
+    name: Push External Images to Charmhub
     runs-on: ubuntu:latest
     needs: [ lookup-external-images ]
     if: ${{ needs.lookup-external-images.outputs.images != '[]' }}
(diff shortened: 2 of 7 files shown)


jdkandersson:
Should this have tests and updates to the documentation?

weiiwang01:
> Should this have tests and updates to the documentation?

I have already updated the test, let me update the document.

.github/workflows/publish_charm.yaml
```diff
@@ -63,6 +63,53 @@ jobs:
...
+    outputs:
+      images: ${{ env.IMAGES }}
+    steps:
+      - uses: actions/checkout@v3
+      - if: inputs.working-directory != './'
+        name: Change directory
+        run: |
+          TEMP_DIR=$(mktemp -d)
+          cp -rp ${{inputs.working-directory}}/. $TEMP_DIR
+          rm -rf .* * || :
```
merkata, line 78: that's like `rm -rf .* * || true`, right?

merkata (dismissed):
Looks good!

.github/workflows/publish_charm.yaml
weiiwang01: Yes, `:` is like a no-op in shell.

.github/workflows/publish_charm.yaml
```diff
@@ -63,6 +63,53 @@ jobs:
...
+          cp -rp $TEMP_DIR/. .
+          rm -rf $TEMP_DIR
```
gregory-schiano, line 103: I'm not sure to understand what this is doing, can you elaborate a bit ?

.github/workflows/publish_charm.yaml
```diff
@@ -63,6 +63,53 @@ jobs:
...
+      - name: Extract External Images
+        run: |
+          IMAGES=$(yq -c '.resources | to_entries | map(select(.value.type == "oci-image" and .value."upstream-source")) | map(.key)' metadata.yaml)
+          echo "External Images: $IMAGES"
+          echo "IMAGES=$IMAGES" >> $GITHUB_ENV
+  publish-external-images:
+    name: Push Externam Images to Charmhub
```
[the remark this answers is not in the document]
weiiwang01: Updated, thanks!

.github/workflows/publish_charm.yaml
weiiwang01: This mostly for test purpose only. It was used when the required working directory is not the project root.

merkata (dismissed):

.github/workflows/build_rocks.yaml
```diff
@@ -79,7 +79,7 @@ jobs:
       - name: Build rock as tarball
         if: ${{ github.event.pull_request.head.repo.fork }}
         run: |
-          skopeo copy oci-archive:$(ls "${{ matrix.path }}"/*.rock) docker-archive:./$IMAGE_NAME.tar:$IMAGE_NAME:${{ github.run_id }}
+          skopeo copy oci-archive:$(ls "${{ matrix.path }}"/*.rock) docker-archive:./$IMAGE_NAME.tar:localhost:32000/$IMAGE_NAME:latest
```
arturo-seijas, line 82: We are using the run_id instead of `latest` to prevent clashed when multiple runs occurs simultaneously. Note that those images might be shared using the registry.

arturo-seijas (changes requested):

.github/workflows/build_rocks.yaml
weiiwang01: This is for the forked situation which uploads the OCI image as an artifact instead of using the registry. But I will just update the tag to use `run_id` to be consistent.

arturo-seijas (approved):
LGTM

.github/workflows/publish_charm.yaml
weiiwang01: Changing the working directory only affect following command in the same step. In the next step, everything will be reset.

.github/workflows/publish_charm.yaml
```diff
@@ -63,6 +63,53 @@ jobs:
...
+          rm -rf $TEMP_DIR
+      - name: Publish image
+        env:
+          CHARMCRAFT_AUTH:  ${{ secrets.CHARMHUB_TOKEN }}
+        run: |
+          image=${{ matrix.image }}
+          charmName=$(yq '.name' metadata.yaml)
+          docker pull $image
+          imageId=$(docker images $image --format "{{.ID}}")
+          charmcraft upload-resource $charmName ${{ matrix.image }}-image --image=$imageId --verbosity=trace
```
amandahla, line 112: I'm not familiar if there is some kind of mechanism to prevent that: usually, external images are always the same. Will the upload recognize that or every time will create a new revision even if there is the same image already uploaded?

amandahla:
Setting upload-image to true wouldn't make charming actions upload the upstream-resource too?

.github/workflows/publish_charm.yaml
weiiwang01: I just tested on my local machine, CharmHub will create a new revision for the same image.
But I think this is okay since the image is stored by sha256 key, so the duplicate image will not cause more storage. Revision is just an entry in the database.

.github/workflows/publish_charm.yaml
gregory-schiano: Yes I get it, but then why don't you change the folder in the same step as the upload-resource command ?

.github/workflows/publish_charm.yaml
gregory-schiano: And there's anyway no way easy way to get the checksump of already uploaded OCI images to avoid reuploading the same image.

Do we want to keep the `--verbosity=trace` ?

PR #147, opened 2023-06-15
fix(ISD-852): Remove duplicate chaos action build

merkata:
This fix removes calling github-chaos-actions twice, as the image responsible for the chaos experiments gets built twice. The logic in the `entrypoint.sh` can install litmus and run the experiments in one pass, thus saving build time and not needing to have two steps, one for setup and one for actually running the tests.

In addition, this PR addresses ISD-849 with pointing to a fix branch that installs the `oci-cli` python package the recommended way (via `pipx`) and the best practive way (via version pinning of the package).


arturo-seijas (approved):
LGTM

weiiwang01 (approved):

PR #145, opened 2023-05-31
Reuse images from pr checks

arturo-seijas:

PR #148, opened 2023-06-15
Eliminate the fixed model from lxd integration tests

gtrkiller:
Having the model "testing" as the fixed model name causes problems with the integration tests when lxd is the provider. Since for this tests it is not useful to have a fixed model name, I'd attempt to remove it for lxd in order for the tests to pass.

### DIFF
--- a/.github/workflows/integration_test_run.yaml
+++ b/.github/workflows/integration_test_run.yaml
@@ -230,52 +230,39 @@ jobs:
       - name: Pre-run script
         if: ${{ inputs.pre-run-script != '' }}
         run: bash -xe ${{ inputs.pre-run-script }}
+      - name: Integration tests variable setting
+        working-directory: ${{ inputs.working-directory }}
+        run: |
+          echo "CHARM_NAME=$([ -f metadata.yaml ] && yq '.name' metadata.yaml || echo UNKNOWN)" >> $GITHUB_ENV
+          args=""
+          for image_name in $(echo '${{ inputs.images }}' | jq -cr '.[]'); do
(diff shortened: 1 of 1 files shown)


arturo-seijas (dismissed):
LGTM

.github/workflows/integration_test_run.yaml
```diff
@@ -252,6 +253,31 @@ jobs:
...
+
+          series=""
+          if [ ! -z ${{ matrix.series }} ]; then
+            series="--series ${{ matrix.series }}"
+          fi
+          module=""
+          if [ ! -z ${{ matrix.modules }} ]; then
+            module="-k ${{ matrix.modules }}"
+          fi
+          tox -e integration -- --keep-models $series $module $args ${{ inputs.extra-arguments }}
```
gregory-schiano, line 278: Isn't there a way to do a bit of factoring between this step and the previous one as 90% of the run script is similar ?
Like having a models_arg output for both steps
And a new step that does all the args/module/series and executes tox using the output of the previous step

gtrkiller: Not that I know of, but I'll take a look and ask around

arturo-seijas: I can do that as part of #145 if that works for you. I'll have to resolve the conflict that will arise as part of this anyway

arturo-seijas (dismissed):
Nice!

weiiwang01 (dismissed):

arturo-seijas (approved):

amandahla (approved):

PR #159, opened 2023-07-05
add cancelled checks and timeouts for required status checks

jdkandersson:
Based on https://github.com/orgs/community/discussions/26303 it seems that jobs with `always()` in the `if` can't be cancelled. Adding a cancelled check and a timeout for the short running jobs that check for the valid status

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -267,7 +267,8 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - integration-test
-    if: always()
+    if: always() && !cancelled()
+    timeout-minutes: 5
     steps:
       - run: |
           [ '${{ needs.integration-test.result }}' = 'success' ] || (echo integration-test failed && false)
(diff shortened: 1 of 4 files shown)


yanksyoon (approved):
LGTM!

gregory-schiano (approved):
LGTM

.github/workflows/integration_test.yaml
```diff
@@ -267,7 +267,8 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - integration-test
-    if: always()
+    if: always() && !cancelled()
+    timeout-minutes: 5
```
arturo-seijas, line 271: Am I missing what this timeout means? Because 5 min doesn't seem to suffice for most cases

jdkandersson: Added the timeout just for the check jobs which just spin up and run a few bash commands. It should be safe, if they are stuck for some reason this will cancel them much sooner

arturo-seijas (approved):

PR #161, opened 2023-07-10
Rockcraft and Charmcraft cache

weiiwang01:


.trivyignore
```diff
@@ -0,0 +1,24 @@
+# https://avd.aquasec.com/nvd/cve-2021-33026
+# Flask-Caching - disputed vurlnerability
+CVE-2021-33026
+# ignore CVEs that are resolved in Indico 3.2.2 until we can use it
```
arturo-seijas, line 4: Why are these here?

.github/workflows/workflow_rockcraft_cache_test.yaml
```diff
@@ -0,0 +1,27 @@
...
+  pull_request:
+  
+jobs:
+  save-cache:
+    uses: ./.github/workflows/build_rocks.yaml
+    secrets: inherit
+    with:
+      owner: ${{ github.repository_owner }}
+      registry: ghcr.io
+      working-directory: "tests/workflows/integration/test-indico-rock/"
```
arturo-seijas, line 16: Is it possible to checkout this repository instead of copying the files?

weiiwang01: That's kind difficult since the `build_rocks` is a reusable workflow instead of an action.

weiiwang01: Yes, but this workflow job doesn't have steps.

.trivyignore
```diff
@@ -0,0 +1,24 @@
...
+CVE-2022-23491
+CVE-2023-0286
+CVE-2023-23931
+CVE-2023-30861
+CVE-2023-28117
+GHSA-39hc-v87j-747x
+CVE-2023-24816
+CVE-2022-40897
+CVE-2022-40898
+# ignore pebble CVEs until https://github.com/canonical/pebble/pull/194 is merged
```
amandahla, line 19: The 194 PR was merged so the CVEs might be removed from here.

.github/workflows/integration_test_run.yaml
```diff
@@ -196,6 +196,33 @@ jobs:
...
+          CHARMCRAFT_BASE="$(yq -o=j -I=0 '.bases | sort_keys(..)' ./charmcraft.yaml)"
+          CHARMCRAFT_BASE=$(printf %s "$CHARMCRAFT_BASE"|jq -sRr @uri)
+          CHARMCRAFT_CACHE_KEY_BASE="charmcraft-cache?name=$CHARM_NAME&bases=$CHARMCRAFT_BASE"
+          echo "CHARMCRAFT_CACHE_KEY=$CHARMCRAFT_CACHE_KEY_BASE&date=$(date +%Y-%m-%d)" >> $GITHUB_ENV
+          echo 'CHARMCRAFT_CACHE_ALT_KEYS<<EOF' >> $GITHUB_ENV 
+          for d in {1..2}
+            do echo "$CHARMCRAFT_CACHE_KEY_BASE&date=$(date -d"-$d days" +%Y-%m-%d)" >> $GITHUB_ENV
+          done
+          echo 'EOF' >> $GITHUB_ENV
+      - name: Restore charmcraft container cache
```
amandahla, line 211: Out of curiosity: Would be possible to have a situation where we might prefer to not use the cache (charmcraft and rockcraft images)?

weiiwang01: The normal integration test workflows only consume caches, there will be two more workflows that build caches. So if certain charm doesn't want to use caches, it can just simply not setup the cache building workflows.

.trivyignore
weiiwang01: I think the entire file might be unnecessary, I will remove this shortly, thanks!

arturo-seijas (commented):
You are changing the image tags, so I don't think the publish workflow will be able to find those. We were using the run_id for this, so that they can be directly linked to a build. See https://github.com/canonical/operator-workflows/blob/main/.github/workflows/publish_charm.yaml#L166

weiiwang01:
> You are changing the image tags, so I don't think the publish workflow will be able to find those. We were using the run_id for this, so that they can be directly linked to a build. See https://github.com/canonical/operator-workflows/blob/main/.github/workflows/publish_charm.yaml#L166

Yes, missed that, sorry about that. I think the solution is that we will just run the image build inside the publish workflow, and since it's cached, it should run relatively quick.

.github/workflows/build_charmcraft_cache.yaml
```diff
@@ -0,0 +1,66 @@
...
+          CHARMCRAFT_CACHE_KEY_BASE="charmcraft-cache?name=$CHARM_NAME&bases=$CHARMCRAFT_BASE"
+          echo "CHARMCRAFT_CACHE_KEY=$CHARMCRAFT_CACHE_KEY_BASE&date=$(date +%Y-%m-%d)" >> $GITHUB_ENV
+          echo 'CHARMCRAFT_CACHE_ALT_KEYS<<EOF' >> $GITHUB_ENV 
+          for d in {1..2}
+            do echo "$CHARMCRAFT_CACHE_KEY_BASE&date=$(date -d"-$d days" +%Y-%m-%d)" >> $GITHUB_ENV
+          done
+          echo 'EOF' >> $GITHUB_ENV
+      - name: Build Charm
+        run: |
+          charmcraft pack
```
arturo-seijas, line 38: If we build the charm here, to avoid packing it again during integration testing, I suggest we upload the artifact here and download it later, packing once to avoid dependencies changing between builds

weiiwang01: The problem is that, one charmcraft takes about 1 GB space, and we only has 10 GB per repository. So we can not save a cache for each pull request, all pull request has to share the cache build on the main branch.

arturo-seijas: The issue with that is that there's no guarantee that the image in CI is the same as the image built during publish

.github/workflows/workflow_rockcraft_cache_test.yaml
arturo-seijas: Noted

.github/workflows/build_charmcraft_cache.yaml
arturo-seijas: I think I didn't explain myself very well. I meant the .charm file resulting from the pack that we are already uploading as an artifact during integration testing

arturo-seijas:
> > You are changing the image tags, so I don't think the publish workflow will be able to find those. We were using the run_id for this, so that they can be directly linked to a build. See https://github.com/canonical/operator-workflows/blob/main/.github/workflows/publish_charm.yaml#L166
> 
> Yes, missed that, sorry about that. I think the solution is that we will just run the image build inside the publish workflow, and since it's cached, it should run relatively quick.

The issue with that is that there's no guarantee that the image in CI is the same as the image built during publish

.github/workflows/build_charmcraft_cache.yaml
weiiwang01: Okay, I will come up with a solution that reuse the charm from the CI workflows in the publish workflow.

weiiwang01:
> > > You are changing the image tags, so I don't think the publish workflow will be able to find those. We were using the run_id for this, so that they can be directly linked to a build. See https://github.com/canonical/operator-workflows/blob/main/.github/workflows/publish_charm.yaml#L166
> > 
> > 
> > Yes, missed that, sorry about that. I think the solution is that we will just run the image build inside the publish workflow, and since it's cached, it should run relatively quick.
> 
> The issue with that is that there's no guarantee that the image in CI is the same as the image built during publish

I think the reproducible build should be guaranteed by rockcraft? Let's see if I can reuse the logic of charm reusing for the image.

gregory-schiano:
> > > > You are changing the image tags, so I don't think the publish workflow will be able to find those. We were using the run_id for this, so that they can be directly linked to a build. See https://github.com/canonical/operator-workflows/blob/main/.github/workflows/publish_charm.yaml#L166
> > > 
> > > 
> > > Yes, missed that, sorry about that. I think the solution is that we will just run the image build inside the publish workflow, and since it's cached, it should run relatively quick.
> > 
> > 
> > The issue with that is that there's no guarantee that the image in CI is the same as the image built during publish
> 
> I think the reproducible build should be guaranteed by rockcraft? Let's see if I can reuse the logic of charm reusing for the image.

Are you sure about that ?
I mean take pip for example, unless you run pip freeze to get all the transient dependencies, if rockcraft re-executes this step you don't have a reproducable build.
I don't know if rockcraft does "part" caching (or layer caching)

yanksyoon (dismissed):
LGTM! A few questions asked :D Thanks for saving a lot of time operator workflows time Weii!

weiiwang01:
> > > > > You are changing the image tags, so I don't think the publish workflow will be able to find those. We were using the run_id for this, so that they can be directly linked to a build. See https://github.com/canonical/operator-workflows/blob/main/.github/workflows/publish_charm.yaml#L166
> > > > 
> > > > 
> > > > Yes, missed that, sorry about that. I think the solution is that we will just run the image build inside the publish workflow, and since it's cached, it should run relatively quick.
> > > 
> > > 
> > > The issue with that is that there's no guarantee that the image in CI is the same as the image built during publish
> > 
> > 
> > I think the reproducible build should be guaranteed by rockcraft? Let's see if I can reuse the logic of charm reusing for the image.
> 
> Are you sure about that ? I mean take pip for example, unless you run pip freeze to get all the transient dependencies, if rockcraft re-executes this step you don't have a reproducable build. I don't know if rockcraft does "part" caching (or layer caching)

I updated the integration test workflow and uploading the image tags used in the integration test. Now the publish workflow should use the same images and charm as the integration test workflow.

.github/workflows/build_rocks.yaml
```diff
@@ -25,31 +25,57 @@ on:
         type: string
         description: The working directory for jobs
         default: "./"
+      cache-action:
```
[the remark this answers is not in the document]
weiiwang01: For rockcraft, there are two different caches. First is the rock cache, which will cache the rock file if the source files didn't change. And if the rock cache missed, there are another layer of cache that cache the rockcraft container to speed up rockcraft pack.

And yes, any project want to use the cache feature must run the cache build workflow on schedule, or the cache feature will be disabled.

.github/workflows/build_charmcraft_cache.yaml
weiiwang01: Updated, the charm and images are now reused in the publish workflow.

.github/workflows/publish_charm.yaml
```diff
@@ -152,21 +149,32 @@ jobs:
...
+      - name: Get charm name
+        id: get-charm-name
+        run: echo "CHARM_NAME=$(yq -e '.name' metadata.yaml)" >> $GITHUB_ENV
       - name: Log in to the Container registry
         uses: docker/login-action@v2
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Download images artifact
```
arturo-seijas, line 161: I assume this is for testing purposes, am I right?

weiiwang01: The integration workflow will upload the image tags used in the workflow as an artifact, and in the publish workflow, the file contains images tags will be downloaded in order to use the same image as the integration test. Images are not in the artifact, only the names.

arturo-seijas: I see. But the artifact won't be available on merge unless you download it using the API. Am I missing something here?

weiiwang01: The image is pushed to the GitHub container registry, so we only need the name to pull the image from Github container registry.

jdkandersson:
This PR actually seems to go some way towards one of our roadmap items around frequently updating the images. It seems that we are essentially keeping the cache hot by periodically building rockcraft images. Could part of this (or a follow up PR) also be that we add a workflow to take that cached image and publish it to charmhub?

.github/workflows/integration_test_run.yaml
jdkandersson, line 1: Something that came to mind, perhaps not for this PR but a follow up PR/ future cycle: we are embedding a lot of complex logic in these workflow files. It would be good to have a Python library which implements most of what we need here so that we can use it in the workflow and be able to test that library

weiiwang01: Yes, we have plans to rewrite the complex login in the operator-workflows

weiiwang01:
> This PR actually seems to go some way towards one of our roadmap items around frequently updating the images. It seems that we are essentially keeping the cache hot by periodically building rockcraft images. Could part of this (or a follow up PR) also be that we add a workflow to take that cached image and publish it to charmhub periodically (e.g., once a day)?

I am not sure if it's possible to update resources in a charm revision without published another charm revision. But other than that, it will be simple to implement.

arturo-seijas:
> > This PR actually seems to go some way towards one of our roadmap items around frequently updating the images. It seems that we are essentially keeping the cache hot by periodically building rockcraft images. Could part of this (or a follow up PR) also be that we add a workflow to take that cached image and publish it to charmhub periodically (e.g., once a day)?
> 
> I am not sure if it's possible to update resources in a charm revision without published another charm revision. But other than that, it will be simple to implement.

Even if it is, we'd have a revision that is mutable. That is, the binaries linked to a revision could transparently change for the people operating the charm. That seems to me like a really big issue

jdkandersson (dismissed):

.github/workflows/build_rockcraft_cache.yaml
```diff
@@ -0,0 +1,31 @@
...
+  build-rockcraft-cache:
+    name: Build rock
+    uses: ./.github/workflows/build_rocks.yaml
+    with:
+      owner: ${{ github.repository_owner }}
+      registry: ghcr.io
+      runs-on: ${{ inputs.runs-on }}
+      trivy-image-config: ${{ inputs.trivy-image-config }}
+      working-directory: ${{ inputs.working-directory }}
+      cache-action: save
```
arturo-seijas, line 31: Minor thigs, but there's a missing blank line here

.github/workflows/publish_charm.yaml
arturo-seijas: Yes, I get you, but this artifact won't be available on merge. You'll need to follow an approach like https://github.com/canonical/operator-workflows/blob/main/.github/workflows/publish_charm.yaml#L188

.github/workflows/integration_test_run.yaml
```diff
@@ -318,3 +350,15 @@ jobs:
           target: ${{ inputs.zap-target-protocol }}://${{ env.ZAP_TARGET }}:${{ inputs.zap-target-port }}/
           cmd_options: ${{ inputs.zap-cmd-options }}
           rules_file_name: ${{ inputs.zap-rules-file-name }}
+      - name: Save image names
```
arturo-seijas, line 353: Can you move these two steps up to the integration_test.yaml to prevent this to be uploaded multiple times? I know this kind of approach caused issues in the past when there are parallel executions

.github/workflows/build_charmcraft_cache.yaml
arturo-seijas: We are also packing in the integration_test.yaml. Should the `charmcraft pack` be removed from there?

.github/workflows/publish_charm.yaml
weiiwang01: Gotcha, adding another step for downloading the images artifact with gh command, thanks!

.github/workflows/integration_test_run.yaml
weiiwang01: Moved, thanks!

jdkandersson (dismissed):

arturo-seijas (dismissed):
Nice! Looks amazing

jdkandersson (approved):

arturo-seijas (approved):

PR #162, opened 2023-07-19
Build charm and enable reusing it during integration testing

arturo-seijas:
* Avoids pushing multiple artifacts when integration tests run in parallel
* Speeds up the pipeline by increasing parallelization
* Speeds up reruns on failed integration tests
* The built artifact is always the tested one

To use it within your tests, just fetch the `charm-file` arg now passed when running the tests
```
def pytest_addoption(parser):
    """Add test arguments."""
    parser.addoption("--charm-file", action="store")
```
```
charm = pytestconfig.getoption("--charm-file")
application = await ops_test.model.deploy(
    charm,
    application_name=app_name,
    series="focal",
)
```

### DIFF
--- a/tests/workflows/integration/test-upload-charm/tests/integration/test_charm.py
+++ b/tests/workflows/integration/test-upload-charm/tests/integration/test_charm.py
@@ -15,7 +15,7 @@ async def test_build_and_deploy(ops_test: OpsTest, pytestconfig):
     """
     app_name = "test"
     assert ops_test.model
-    charm = await ops_test.build_charm(".")
+    charm = pytestconfig.getoption("--charm-file")
     resources = {"test-image": pytestconfig.getoption("--test-image")}
 
     await asyncio.gather(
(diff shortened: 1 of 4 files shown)


tests/workflows/integration/test-upload-charm/tests/integration/test_charm.py
```diff
@@ -15,7 +15,7 @@ async def test_build_and_deploy(ops_test: OpsTest, pytestconfig):
     """
     app_name = "test"
     assert ops_test.model
-    charm = await ops_test.build_charm(".")
+    charm = pytestconfig.getoption("--charm-file")
```
yanksyoon, line 18: Minor nitpick, should this be better suited in conftest?

yanksyoon (approved):
LGTM!

tests/workflows/integration/test-upload-charm/tests/integration/test_charm.py
arturo-seijas: The other option is also fetched this way. I don't see why not changing it, though

gtrkiller (approved):
LGTM

PR #163, opened 2023-07-20
Build charm and enable reusing it during integration testing

arturo-seijas:
Follow-up #162 

To use it within your tests, just fetch the `charm-file` arg now passed when running the tests
```
def pytest_addoption(parser):
    """Add test arguments."""
    parser.addoption("--charm-file", action="store")
```
```
charm = pytestconfig.getoption("--charm-file")
application = await ops_test.model.deploy(
    f"./{charm}",
    application_name=app_name,
    series="focal",
)
```

### DIFF
--- a/.github/workflows/integration_test_run.yaml
+++ b/.github/workflows/integration_test_run.yaml
@@ -223,7 +223,7 @@ jobs:
             args="${args} --${image_name}-image ${{ inputs.registry }}/${{ inputs.owner }}/${image_name}:${{ github.run_id }}"
           done
           if [ ! -z ${{ inputs.charm-file }} ]; then
-            args="${args} --charm-file ./${{ inputs.charm-file }}"
+            args="${args} --charm-file ${{ inputs.charm-file }}"
           fi
           echo "ARGS=$args" >> $GITHUB_ENV
           series=""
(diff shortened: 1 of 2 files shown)


amandahla (approved):

mthaddon:
It seems like you're including documentation in the PR description. Should we update the README instead/as well?

arturo-seijas:
> It seems like you're including documentation in the PR description. Should we update the README instead/as well?

We don't havve any docs on how to write the tests. Perhaps we can include those in the readme, but I think that's out of scope ATM

jdkandersson:
> > It seems like you're including documentation in the PR description. Should we update the README instead/as well?
> 
> We don't havve any docs on how to write the tests. Perhaps we can include those in the readme, but I think that's out of scope ATM

I'm not following, don't we describe how to use the operator workflows in the readme? This is part of how to use it in pytest?

arturo-seijas:
> > > It seems like you're including documentation in the PR description. Should we update the README instead/as well?
> > 
> > 
> > We don't havve any docs on how to write the tests. Perhaps we can include those in the readme, but I think that's out of scope ATM
> 
> I'm not following, don't we describe how to use the operator workflows in the readme? This is part of how to use it in pytest?

We do, but there's nothing about how to write the tests. I'll open a PR with some addional documentation

mthaddon (approved):
LGTM, given the separate PR for doc changes

PR #165, opened 2023-07-20
Do not pack the charm if the metadata file is not present

arturo-seijas:
Do not pack the charm if the metadata file is not present to support repositories not containing charms

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -225,8 +225,8 @@ jobs:
       - uses: actions/checkout@v3
       - uses: canonical/setup-lxd@v0.1.1
       - name: Get charm name
-      run: |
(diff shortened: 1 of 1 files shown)


gregory-schiano:
Important note ! Relying on metadata.yaml presence is a bad move since we're merging metadata and config into charmcraft.yaml, therefore metadata.yaml will become optional !

arturo-seijas:
> Important note ! Relying on metadata.yaml presence is a bad move since we're merging metadata and config into charmcraft.yaml, therefore metadata.yaml will become optional !

Thanks for the heads up. We'll have to support both formats, but there are already a lot of places where the metadata info is parsed, so it will require a significant work

jdkandersson (dismissed):

gtrkiller (dismissed):
LGTM

jdkandersson (approved):

mthaddon (approved):
LGTM

PR #177, opened 2023-07-28
Add pre-run-script to the test workflow

weiiwang01:


arturo-seijas (approved):
LGTM

cbartz (approved):
LGTM

.github/workflows/test.yaml
```diff
@@ -14,6 +14,9 @@ on:
         type: boolean
         description: Whether to use self-hosted runners to run the jobs.
         default: true
+      pre-run-script:
+        description: Path to the bash script to be run before the integration tests
```
jdkandersson, line 18: Should this say "integration" test? Also, should this be documented in the readme?

weiiwang01: I think this is copied from the input definition of the integration test workflow. I will fix the in the next pull request, thanks!

PR #201, opened 2023-09-18
Add inputs for unit test for selection the self-hosted runner

yhaliaw:
Applicable spec: NA

### Overview

Add input to unit test workflow for selecting self-hosted runners by label.

### Rationale

Needed for using different version of self-hosted runners.

### Workflow Changes

One new input for unit test workflow added.

### Checklist

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -27,21 +27,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  runs-on-labels:
-    runs-on: ['self-hosted', 'small']
-    outputs:
-      labels: ${{ steps.labels.outputs.LABELS }}
-    steps:
-      - id: labels
-        run: |
-          if ${{ inputs.self-hosted-runner }}; then
-            echo "LABELS=['self-hosted', '${{ inputs.self-hosted-runner-label }}']" >> "$GITHUB_OUTPUT"
-          else
-            echo "LABELS=['ubuntu-22.04']" >> "$GITHUB_OUTPUT"
-          fi
   inclusive-naming-check:
     name: Inclusive naming
-    needs: runs-on-labels
     runs-on: >-
       ${{
         inputs.self-hosted-runner &&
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yaml
```diff
@@ -23,9 +27,22 @@ concurrency:
...
+    outputs:
+      labels: ${{ steps.labels.outputs.LABELS }}
+    steps:
+      - id: labels
+        run: |
+          if ${{ inputs.self-hosted-runner }}; then
+            echo "LABELS=['self-hosted', '${{ inputs.self-hosted-runner-label }}']" >> "$GITHUB_OUTPUT"
+          else
+            echo "LABELS=['ubuntu-22.04']" >> "$GITHUB_OUTPUT"
+          fi
```
jdkandersson, line 41: I'm curios why this is needed? Is there some limitation of what can be put into `runs-on`? Perhaps it is that it needs to be an array?

jdkandersson (dismissed):

.github/workflows/test.yaml
```diff
@@ -23,9 +27,22 @@ concurrency:
...
   inclusive-naming-check:
     name: Inclusive naming
-    runs-on: ${{ inputs.self-hosted-runner && fromJson('[''self-hosted'', ''large'']') || 'ubuntu-22.04' }}
```
weiiwang01, line 28: How about something this, so we can save the `runs-on-labels` job,
```yaml
runs-on: >-
  ${{
    inputs.self-hosted-runner &&
    fromJson(format('[''self-hosted'', ''{0}'']',  inputs.self-hosted-runner-label)) ||
    'ubuntu-22.04'
    }}
```

jdkandersson (dismissed):

.github/workflows/test.yaml
jdkandersson: I'm guessing this isn't needed anymore now since the other way that Weii suggested seems to work?

jdkandersson (approved):

weiiwang01 (approved):

PR #205, opened 2023-10-13
Remove Android SDK to free up space in root

weiiwang01:
### Overview

The GitHub hosted runners currently only have 18GB of free space in the root directory, which may be insufficient for some integration tests. Removing Android SDK to reclaim about 15GB in the root.

### Workflow Changes

None.

### Checklist

### DIFF
--- a/.github/workflows/integration_test_run.yaml
+++ b/.github/workflows/integration_test_run.yaml
@@ -189,6 +189,8 @@ jobs:
     if: ${{ !failure() }}
     steps:
       - uses: actions/checkout@v4.1.0
+      - name: Remove Android SDK
+        run: sudo rm -rf /usr/local/lib/android
       - name: Setup Devstack Swift
         if: ${{ inputs.setup-devstack-swift }}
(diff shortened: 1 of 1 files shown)


mthaddon (approved):

merkata (approved):
Kind of a forceful removal, if it's packaged perhaps we can purge the whole package (in the future)?

weiiwang01:
> Kind of a forceful removal, if it's packaged perhaps we can purge the whole package (in the future)?

It seems like the Android SDK is downloaded directly instead of using a package manager in self-hosted runners https://github.com/actions/runner-images/blob/main/images/linux/scripts/installers/android.sh

merkata:
> > Kind of a forceful removal, if it's packaged perhaps we can purge the whole package (in the future)?
> 
> It seems like the Android SDK is downloaded directly instead of using a package manager in self-hosted runners https://github.com/actions/runner-images/blob/main/images/linux/scripts/installers/android.sh

Thanks for checking, removing it this way seems logical then.

PR #211, opened 2023-11-02
Feat/trivyignore

merkata:
Applicable spec: ISD-084

### Overview

Annotate ignored CVEs that can be removed as they are not present anymore.

### Rationale

Cleanup of unused config.

### Workflow Changes

An additional step in the integration tests build process for rocks will now annotate .trivyignore and tell you if something can be removed.

### Checklist

- [x ] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [x ] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)

image
^ an implementation PoC from wordpress-k8s


.github/workflows/autoremove_trivyignore.yaml
```diff
@@ -0,0 +1,57 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+name: Auto-remove ignored CVEs
+on:
+  workflow_call:
+
+jobs:
+  update-lib:
```
arturo-seijas, line 9: Can you rename this job?

.github/workflows/build_rocks.yaml
```diff
@@ -224,3 +226,23 @@ jobs:
         env:
           TRIVY_USERNAME: ${{ github.actor }}
           TRIVY_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
+      - name: Check trivyignore
+        run: |
+          curl -sfL https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh | sh -s -- -b /usr/local/bin v0.46.0
+          if [ -f ".trivyignore" ]
+          then
+            output=$(trivy image $ROCK_IMAGE --severity HIGH,CRITICAL -q -f json --ignorefile "" | jq -r '.Results[].Vulnerabilities[].VulnerabilityID' 2>/dev/null || echo "No vulnerabilities found")
```
arturo-seijas, line 234: Isn't the severity configurable via config file? If so, we should probably take it into consideration

.github/workflows/autoremove_trivyignore.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4.1.1
+        with:
+          fetch-depth: 0
+      - name: Check trivyignore
+        run: |
+          curl -sfL https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh | sh -s -- -b /usr/local/bin v0.18.3
+          if [ -f ".trivyignore" ]
+          then
+            output=$(trivy fs . -q -f json --ignorefile "" | jq -r '.Results[].Vulnerabilities[].VulnerabilityID' 2>/dev/null || echo "No vulnerabilities found")
```
amandahla, line 26: The [trivyignore](https://aquasecurity.github.io/trivy/v0.46/docs/configuration/filtering/#by-finding-ids) might have more than CVEs. Would be possible to compare the output running without trivyignore and with trivyignore and then create a PR to check the difference?

merkata: I'm resolving this as we're going to annotate the trivyignore against CVEs that are outdated, so comparing output of CVEs (without trivyignore) against these present in the trivyignore and annotate the file. No PR will be created, we will be using the approach that we have with hadolint (annotations). Outdated CVEs are a chore, so that's a good solution imo.

.github/workflows/autoremove_trivyignore.yaml
merkata: I will remove this job, we will be annotating the trivyignore based on CVEs found in the rock and that job is no longer needed, thanks for spotting anyway!

.github/workflows/build_rocks.yaml
merkata: The Run Github Trivy Image Action has this configured, so I used it as well:
```
354094af (arturo-seijas                2023-02-20 16:21:56 +0100 225)           severity: 'CRITICAL,HIGH'
```
I can remove it though, as we can scan for all severities here, it won't matter - if they are in the trivyignore, we'll annotate them.

.github/workflows/build_rocks.yaml
```diff
@@ -224,3 +226,23 @@ jobs:
         env:
           TRIVY_USERNAME: ${{ github.actor }}
           TRIVY_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
+      - name: Check trivyignore
```
jdkandersson, line 229: Do you think we should write this in nodeJS given that we want to start going down that path throughout this cycle?

merkata: I prefer to land this as is if you're fine with it - I think we need some initial setup around rewriting in javascript and also some research. The story itself started before we decided on the scope of rewriting ops workflows in javascript.

.github/workflows/build_rocks.yaml
arturo-seijas: I'd rather remove them, keeping maintainability in mind

amandahla (approved):
LGTM

jdkandersson (approved):

.github/workflows/build_rocks.yaml
merkata: I'll open a story about this in the backlog, I want to have a quick alignment on how we configure both actions, the one introduced here is more a no-op, I'm more interested in seeing about the other.

merkata: Resolving to unblock the PR, though if you mind, please re-open it, that's fine with me.

PR #224, opened 2023-12-11
Integration/multi charm artifacts

addyess:
### Overview

Support building multiple charms from the same repository and test with a single integration build.

### Rationale

Some charm repos may hold multiple charms and need to have their integration testing run with a single commit change.  It's possible to define multiple integration-tests with different working-directories -- but without these changes its not possible to build both charms and run under one integration tests run. 

These changes allow the charms to build, then artifact the charm files in parallel with a matrix run of `charmcraft pack` for each of those sub-directories, then share a single integration-test-run for them all together

### Workflow Changes

The major shift is to store all charm file artifacts in an artifact named `charm-artifact`.  This is a file store which multiple files can be pushed to in each of the matrix runs, then pulled by the single integration-run.  This replaces the previous technique to store each charm in its own `${{ charm.name}}-image` artifact.

### Checklist

### DIFF
--- a/.github/workflows/build_charms.yaml
+++ b/.github/workflows/build_charms.yaml
@@ -30,16 +30,16 @@ jobs:
       - name: Find charm paths
         working-directory: ${{ inputs.working-directory }}
         run: |
-          echo "CHARM_PATHS=$(find . -name metadata.yaml 2> /dev/null |
+          echo "CHARM_PATHS=$(find * -type f -name metadata.yaml 2> /dev/null |
           sed 's:[^/]*$::' |
           jq -Rsc '. / "\n" - [""]')" >> $GITHUB_ENV
       - name: Extract charm names
         working-directory: ${{ inputs.working-directory }}
         run: |
           echo "CHARM_NAMES=$(for path in $(jq -cr '.[]' ); do
-            echo "{\"name\":\"$(yq '.name' $path/metadata.yaml)\", \"path\":\"$path\}"
+            echo {\"name\":\"$(yq '.name' $path/metadata.yaml)\", \"path\":\"$path\"}
           done <<< $CHARM_PATHS |
-          jq -Rsc '.')" >> $GITHUB_ENV
+          jq -sc '.')" >> $GITHUB_ENV
   build-charms:
     name: Build and push charms
     runs-on: ${{ inputs.runs-on }}
@@ -62,5 +62,5 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: charm-artifacts
-          path: ${{ inputs.working-directory }}/${{ matrix.path }}/${{ env.CHARM_FILE }}
+          path: ${{ inputs.working-directory }}/${{ matrix.path }}${{ env.CHARM_FILE }}
           if-no-files-found: error
--- a/.github/workflows/build_charms.yaml
+++ b/.github/workflows/build_charms.yaml
@@ -30,7 +30,7 @@ jobs:
       - name: Find charm paths
         working-directory: ${{ inputs.working-directory }}
         run: |
-          echo "CHARM_PATHS=$(find . -name metadata.yaml 2> /dev/null) |
+          echo "CHARM_PATHS=$(find . -name metadata.yaml 2> /dev/null |
           sed 's:[^/]*$::' |
           jq -Rsc '. / "\n" - [""]')" >> $GITHUB_ENV
       - name: Extract charm names
--- a/.github/workflows/build_charms.yaml
+++ b/.github/workflows/build_charms.yaml
@@ -1,0 +1,66 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+name: Build Charms
+
+on:
+  workflow_call:
+    inputs:
+      runs-on:
+        type: string
+        description: Image runner for building the images
+        default: ubuntu-22.04
+      working-directory:
+        type: string
+        description: The working directory for jobs
+        default: "./"
+    outputs:
+      charms:
+        description: List of charms built
+        value: ${{ jobs.get-charms.outputs.charms }}
+
+jobs:
+  get-charm-names:
+    name: Get charms
+    runs-on: ${{ inputs.runs-on }}
+    outputs:
+      charms: ${{ env.CHARM_NAMES }}
+    steps:
+      - uses: actions/checkout@v4.1.1
+      - name: Find charm paths
+        working-directory: ${{ inputs.working-directory }}
+        run: |
+          echo "CHARM_PATHS=$(find . -name metadata.yaml 2> /dev/null) |
+          sed 's:[^/]*$::' |
+          jq -Rsc '. / "\n" - [""]')" >> $GITHUB_ENV
+      - name: Extract charm names
+        working-directory: ${{ inputs.working-directory }}
+        run: |
+          echo "CHARM_NAMES=$(for path in $(jq -cr '.[]' ); do
+            echo "{\"name\":\"$(yq '.name' $path/metadata.yaml)\", \"path\":\"$path\}"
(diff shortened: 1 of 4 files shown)


addyess:
@jdkandersson I'm not sure why these failure regarding `build-rocks` report as workflow errors.  

My only guess is that it has something to do from the changes coming from my fork.

.github/workflows/integration_test.yaml
```diff
@@ -265,11 +265,19 @@ jobs:
           name: ${{ env.CHARM_NAME }}-images
           path: ${{ inputs.working-directory }}/${{ env.CHARM_NAME }}-images
           if-no-files-found: error
+  build-multiple-charms:
+    name: Build and push multiple charms out of subdirectories
+    needs: [get-runner-image, build-charm]
+    if: ${{ needs.build-charm.outputs.charm-file == '' && !cancelled() }}
```
arturo-seijas, line 271: Isn't the file always available?

addyess: Not always.  Envision a repo like this:

```
my-multi-charms
└── charms
    ├── charm-a
    │   └── metadata.yaml
    └── charm-b
        └── metadata.yaml
```

if there's not a `metadata.yaml` in the working-directory, then there's no charm-file built

jdkandersson:
> @jdkandersson I'm not sure why these failure regarding `build-rocks` report as workflow errors.
> 
> My only guess is that it has something to do from the changes coming from my fork.

I'm hoping that @arturo-seijas can help debug this one, I'm not as familiar with this part of the workflow

jdkandersson:
The same on here, I believe the failing workflows are not actually related to the PR and instead due to permission issues due to the workflow being from a fork. Do you have a workflow run which demonstrates the feature working please? That would help us get this merged

.github/workflows/integration_test.yaml
arturo-seijas: Understood

.github/workflows/build_charms.yaml
```diff
@@ -0,0 +1,66 @@
...
+          sudo snap install charmcraft --classic --channel latest/stable
+          charmcraft pack -v
+          echo "CHARM_FILE=$(ls ${{matrix.name}}_*.charm)" >> $GITHUB_ENV
+      - name: Upload charm artifact
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: charm-artifacts
+          path: ${{ inputs.working-directory }}/${{ matrix.path }}${{ env.CHARM_FILE }}
+          if-no-files-found: error
```
weiiwang01, line 66: I think artifacts with the same name will overwrite each other? So  I think the matrix here will only save one charm as artifact.

arturo-seijas (commented):
I haven't looked into the failing tests yet but this will probably break at https://github.com/canonical/operator-workflows/blob/main/.github/workflows/publish_charm.yaml#L204

arturo-seijas (commented):
/canonical/self-hosted-runners/run-workflows cfec3ea9187ad90e01d17180b6d61ef584a70391

arturo-seijas:
/canonical/self-hosted-runners/run-workflows cfec3ea9187ad90e01d17180b6d61ef584a70391

.github/workflows/integration_test_run.yaml
```diff
@@ -261,8 +258,15 @@ jobs:
...
         with:
-          name: ${{ env.CHARM_NAME }}-charm
-          path: ${{ inputs.working-directory }}
+          name: charm-artifacts
+      - name: Add charm-file arguments
+        working-directory: ${{ inputs.working-directory }}
+        run: |
+          args=${{ env.ARGS }}
+          for charm in $(ls *.charm); do
+            args="${args} --charm-file=${charm}"
```
weiiwang01, line 267: I think `*.charm` will only list the charm files in the current directory, and `**/*.charm` is to recursively list all charm files? 
Another question is that if we are using the same argument construction `args="${args} --charm-file=${charm}"`, which would produce something like `--charm-file=aaa/aaa.charm --charm-file=bbb/bbb.charm --charm-file=ccc/ccc.charm`, then how does the integration know which charm to use?

cbartz:
/canonical/self-hosted-runners/run-workflows c55f1fbcbc3db8d56f3af6c3a5b0891b71446949

.github/workflows/build_charms.yaml
addyess: correct, artifacts with the same name would clash.

I am attempting to push charms like this into the file store
`aaa_ubuntu-22.04-amd64.charm`
`bbb_ubuntu-22.04-amd64.charm`

if two or more sub-directories had charms whose resultant filenames were the same (ie -- two charm file both with the name `ccc__ubuntu-22.04-amd64.charm`) then only one would be stored in the artifact for the action run.  I'm not entirely sure how to solve that -- even `${{ env.CHARM_NAME }}-charm` doesn't address this so it's no worse than the existing implementation

addyess:
> The same on here, I believe the failing workflows are not actually related to the PR and instead due to permission issues due to the workflow being from a fork. Do you have a workflow run which demonstrates the feature working please? That would help us get this merged

Let me get back to you on this as currently my repo is private and for the time being I've got a workaround in place to build the charms with my own workflows, then push all the charms to both the `${{ env.CHARM_NAME }}-charm` artifact and `UNKNOWN-charm` artifact.  This does the trick of getting the built charms into the right spot for the integration testing without breaking the rest of the workflows

.github/workflows/build_charms.yaml
weiiwang01: Didn't know GitHub workflow artifacts is merged when upload multiple times, thanks for the information!

.github/workflows/integration_test_run.yaml
weiiwang01: Gotcha, if that's the case, I think we can remove the `build-charm` job and related steps, since that's just `build-charms` but with only one charm.

.github/workflows/build_charms.yaml
addyess: https://github.com/actions/upload-artifact#usage

oh sure, workflow artifact uploading is cool.
* individual files
* entire directories
* wildcard pattern uploads
* multiple-paths with file matching exclusions

.github/workflows/integration_test_run.yaml
addyess: yeah, i was trying to think of a way to not break other users of these workflows.  That's why I only ran `build-charms` job if `build-charm` didn't produce a `charm-file`

if there are others depending on these artifacts -- then it could be an issue.  

It's also worth noting i didn't take any care to build rocks or images for the  multiple charms case -- nor did i try to find a way to distinctly manage which charm needs which image. 

I'm not totally sure this PR contains all the right steps needed to manage these complexities. Perhaps it should be closed and I try again with a spec. 

It won't be long to where I need multiarch charms.  This will produce multiple `.charm` files for each supported architecture.  Each of these will need to be published associated with the respective image resources for those architectures.  Again -- another reason for perhaps stepping back and making a spec.

weiiwang01: Oh, yes, that reminds me that the publish workflow actually uses the old charm artifact. I think the entire `operator-workflows` will be rewritten in the future, but I think I am good with the current status. Thanks!

weiiwang01 (approved):

cbartz (commented):
I think we should also have an integration test for the multiple charms use case.

addyess:
Well @weiiwang01 that didn't last very long.  `actions/upload-artifact@v4` has decided to break this feature
> Uploading to the same named Artifact multiple times
https://github.com/actions/upload-artifact/tree/v4#breaking-changes

This won't be a tenable solution anymore nor am i sure quite how to get a matrix job to unify each matrix strategies artifacted outputs.

PR #236, opened 2024-01-16
Enable poetry workflow

jdkandersson:
Applicable spec: ISD115

### Overview

- Install poetry in the unit/integ test runner with the poetry-plugin-export plugin
- Optionally uses the `tox -e pack` command instead of `charmcraft pack` to enable the use of poetry
- Remove `tox.ini`'s dependencies check since they will be located in `pyproject.toml`

### Rationale

Enable workflows with full dependency pinning using poetry.  
PRs using this branch:
- https://github.com/canonical/bind-operator/pull/19 (using poetry)
- https://github.com/canonical/discourse-k8s-operator/pull/250 (using poetry)
- https://github.com/canonical/synapse-operator/pull/223 (using poetry)

### Workflow Changes

The integration test has been changed to optionally use `tox -e pack` command instead of `charmcraft pack`

### Checklist


.github/workflows/integration_test.yaml
```diff
@@ -244,7 +244,8 @@ jobs:
         working-directory: ${{ inputs.working-directory }}
         run: |
           sudo snap install charmcraft --classic --channel latest/stable
-          charmcraft pack -v
+          python -m pip install tox
+          tox -e pack -- -v | charmcraft pack -v
```
merkata, line 248: should be "||", no?

jdkandersson: I think you're right, I have fixed that. I wonder why it worked anyway, I think `|` is bitwise or which probably is fine as well in this case

.github/workflows/test.yaml
```diff
@@ -216,51 +216,6 @@ jobs:
           STDOUT_LOG=$(mktemp --suffix=stdout.log)
           echo STDOUT_LOG=$STDOUT_LOG >> $GITHUB_ENV
           tox --result-json=test-result.json | tee $STDOUT_LOG ; test ${PIPESTATUS[0]} -eq 0
-      - name: Check lint and test stdout
```
jdkandersson, line 219: What is the reason for removing this please?

jdkandersson:
> This is now marked ready for review, but I think the PR description is out of sync (e.g. `Install poetry in the unit/integ test runner with the poetry-plugin-export plugin` is not present).

Good point, perhaps @nrobinaubertin would you like to take over this PR please?

.github/workflows/test.yaml
nrobinaubertin: When using poetry, the dependencies will be handled in the `pyproject.toml` file and this test will always fail

nrobinaubertin:
> > This is now marked ready for review, but I think the PR description is out of sync (e.g. `Install poetry in the unit/integ test runner with the poetry-plugin-export plugin` is not present).
> 
> Good point, perhaps @nrobinaubertin would you like to take over this PR please?

I'll take it over.  
But I did not want to put it in review because of https://github.com/canonical/operator-workflows/pull/236#issuecomment-1965623504

.github/workflows/test.yaml
jdkandersson: Could it somehow be adapted so that we still get the benefits? It is checking here that the expected tools are being run which is quite helpful to make sure that the expected linting is actually being done in a given repo

merkata:
@nrobinaubertin @jdkandersson is this being worked on? I suggest closing it if not and we can reopen it if needed..

cbartz:
@nrobinaubertin @jdkandersson Can we get a status update here, please? Is this still/ or supposed to being worked on? Thanks.

nrobinaubertin:
Closing as not on our roadmap anymore

PR #243, opened 2024-01-22
KU-52 Make the build-rocks job multiarch aware

addyess:
Updates the `build-rocks` job to be aware of the potential for a `rockcraft.yaml` to have multiple platforms.  It doesn't actually try to build those images yet appropriately on the correct architecture, but it makes it possible to distinctly identify those images when the `canonical/craft-actions/rockcraft-pack` grows the ability to build specific architectures

### DIFF
--- a/.github/workflows/build_rocks.yaml
+++ b/.github/workflows/build_rocks.yaml
@@ -29,10 +29,17 @@ on:
         type: string
         description: The cache action can either be "save" or "restore".
         default: restore
+      multiarch-awareness:
+        type: boolean
+        description: Maintain the architecture labels on the container names
+        default: false
     outputs:
       images:
         description: List of images built
         value: ${{ jobs.get-rocks.outputs.images }}
+      rock-metas:
+        description: List of maps featuring the built {path, arch, image}
+        value: ${{ jobs.get-rocks.outputs.rock-metas }}
 
 jobs:
   get-rocks:
@@ -41,6 +48,7 @@ jobs:
     outputs:
       rock-paths: ${{ steps.gen-rock-paths-and-images.outputs.rock-paths }}
       images: "${{ steps.gen-rock-paths-and-images.outputs.images }}"
+      rock-metas: ${{ steps.gen-rock-paths-and-images.outputs.rock-metas }}
     steps:
       - name: Validate inputs
         run: |
@@ -55,24 +63,44 @@ jobs:
         with:
           script: |
             const path = require('path')
-            const workingDir = core.getInput('working-directory')
+            const inputs = ${{ toJSON(inputs) }}
+            const workingDir = inputs['working-directory']
+            const multiarch = inputs['multiarch-awareness']
             const rockcraftGlobber = await glob.create(
               path.join(workingDir, '**/rockcraft.yaml')
             )
             const rockPaths = []
             const images = []
+            const rockMetas = []
+            const defaultArch = 'amd64'
+            core.info(`Multiarch Awareness is ${multiarch ? "on" : "off" }`)
             for (const rockcraftFile of await rockcraftGlobber.glob()) {
               const rockPath = path.relative('.', path.dirname(rockcraftFile))
-              rockPaths.push(rockPath)
               core.info(`found rockcraft.yaml in ${rockPath}`)
               const fileHash = await glob.hashFiles(path.join(rockPath, '**'))
               const rockName = (
                 await exec.getExecOutput('yq', ['.name', rockcraftFile])
               ).stdout.trim()
-              const image = `${{ inputs.registry }}/${{ inputs.owner }}/${rockName}:${fileHash}`
-              core.info(`generate image name: ${image}`)
-              images.push(image)
+              const platforms = (
+                await exec.getExecOutput('yq', ['.platforms | keys', '-o=json', rockcraftFile])
+              ).stdout.trim()
+              if (multiarch && platforms) {
+                const arches = JSON.parse(platforms)
+                for (arch of arches) {
+                  const image = `${{ inputs.registry }}/${{ inputs.owner }}/${rockName}:${fileHash}-${arch}`
+                  core.info(`generate multi-arch image name: ${image}`)
+                  images.push(image)
+                  rockMetas.push({'path': rockPath, 'arch': arch, 'image': image})
+                }
+              } else {
+                const image = `${{ inputs.registry }}/${{ inputs.owner }}/${rockName}:${fileHash}`
+                core.info(`generate image name: ${image}`)
+                images.push(image)
+                rockMetas.push({'path': rockPath, 'arch': defaultArch, 'image': image})
+              }
+              rockPaths.push(rockPath)
             }
+            core.setOutput('rock-metas', JSON.stringify(rockMetas))
             core.setOutput('rock-paths', JSON.stringify(rockPaths))
             core.setOutput('images', JSON.stringify(images))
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build_rocks.yaml
```diff
@@ -29,10 +29,17 @@ on:
         type: string
         description: The cache action can either be "save" or "restore".
         default: restore
+      multiarch-awareness:
+        type: boolean
+        description: Maintain the architecture labels on the container names
+        default: false
```
addyess, line 35: Forces the build to create a separate build for each architecture specified in the `rockcraft.yaml` platforms

.github/workflows/build_rocks.yaml
```diff
@@ -55,24 +63,44 @@ jobs:
         with:
           script: |
             const path = require('path')
-            const workingDir = core.getInput('working-directory')
+            const inputs = ${{ toJSON(inputs) }}
+            const workingDir = inputs['working-directory']
```
addyess, line 67: Adjusting since `core.getInput` doesn't quite work right

https://github.com/actions/toolkit/issues/1124#issuecomment-1305836110

.github/workflows/build_rocks.yaml
```diff
@@ -55,24 +63,44 @@ jobs:
...
+            core.info(`Multiarch Awareness is ${multiarch ? "on" : "off" }`)
             for (const rockcraftFile of await rockcraftGlobber.glob()) {
               const rockPath = path.relative('.', path.dirname(rockcraftFile))
-              rockPaths.push(rockPath)
               core.info(`found rockcraft.yaml in ${rockPath}`)
               const fileHash = await glob.hashFiles(path.join(rockPath, '**'))
-              const rockName = (
-                await exec.getExecOutput('yq', ['.name', rockcraftFile])
+              const [rockName, rockVersion] = (
+                await exec.getExecOutput('yq', ['.name,.version', rockcraftFile])
```
addyess, line 82: parse out the `name` and `version` from the rockcraftFile

.github/workflows/build_rocks.yaml
```diff
@@ -55,24 +63,44 @@ jobs:
...
+              ).stdout.trim().split("\n")
+              const platforms = (
+                await exec.getExecOutput('yq', ['.platforms | keys', '-o=json', rockcraftFile])
               ).stdout.trim()
```
addyess, line 86: pull out the `platforms` list from the rockcraft file as a json list

.github/workflows/build_rocks.yaml
```diff
@@ -55,24 +63,44 @@ jobs:
...
-              core.info(`generate image name: ${image}`)
-              images.push(image)
+              if (multiarch && platforms) {
+                const arches = JSON.parse(platforms)
+                for (arch of arches) {
+                  const image = `${{ inputs.registry }}/${{ inputs.owner }}/${rockName}:${fileHash}-${arch}`
+                  core.info(`generate multi-arch image name: ${image}`)
+                  images.push(image)
+                  rockMetas.push({name: rockName, version: rockVersion, path: rockPath, arch: arch, image: image})
+                }
```
addyess, line 94: when multiarch is engaged and `platforms` are described by `rockcraft.yaml`, push a separate image for each platform with the `arch` appended at the end of the tag

.github/workflows/build_rocks.yaml
```diff
@@ -55,24 +63,44 @@ jobs:
...
+              } else {
+                const image = `${{ inputs.registry }}/${{ inputs.owner }}/${rockName}:${fileHash}`
+                core.info(`generate image name: ${image}`)
+                images.push(image)
+                rockMetas.push({name: rockName, version: rockVersion, path: rockPath, arch: defaultArch, image: image})
```
addyess, line 99: maintain the prior behavior assuming `amd64` as the `defaultArch`

.github/workflows/build_rocks.yaml
```diff
@@ -83,33 +111,39 @@ jobs:
       contents: read
       packages: write
     needs: [get-rocks]
-    if: ${{ needs.get-rocks.outputs.rock-paths != '[]' }}
+    if: ${{ needs.get-rocks.outputs.rock-metas != '[]' }}
     strategy:
       matrix:
-        path: ${{ fromJSON(needs.get-rocks.outputs.rock-paths) }}
+        rock: ${{ fromJSON(needs.get-rocks.outputs.rock-metas) }}
```
addyess, line 117: this is now a list of `metas` -- `[<meta>,<meta>,<meta>]`
a meta is an object that looks like:
```json
{
    name: <rock.name>,
    version: <rock.version>,
    path: <path to rockcraft.yaml>,
    arch:  <rock platform>,
    image: <expected image name>
}
```

weiiwang01:
I think the issue lies with the rockcraft, as it should generate a single multi-architecture rock file (multi-architecture OCI image archive), rather than creating multiple rock files, each for a different architecture. And, this multi-architecture rock file can be uploaded to the container registry as a single multi-architecture image.

addyess:
@weiiwang01, that's not my understanding of how oci-archives work.  

A "multiarch" image doesn't exist.  In order to create a multi-arch image, one must create the images uniquely -- this will result in 2 or more `rock` files each compiled for an architecture.  

1) Each image should be individually pushed to the registry with a tag like `repo/myimage:ver-arch`
2) an oci-manifest is created named `repo/myimage:ver` which references like a pointer to the correct images

These changes allow this job to create the `.rock` files and push them as `repo/myimage:ver-arch` (part 1)

As of now, charmhub does not support uploading a multiarch manifest -- it only supports images.

I understand if you want to reject this PR -- I'll end up building multiarch images to and pushing ghcr.io in other ways if so

weiiwang01:
> @weiiwang01, that's not my understanding of how oci-archives work.
> 
> A "multiarch" image doesn't exist. In order to create a multi-arch image, one must create the images uniquely -- this will result in 2 or more `rock` files each compiled for an architecture.
> 
> 1. Each image should be individually pushed to the registry with a tag like `repo/myimage:ver-arch`
> 2. an oci-manifest is created named `repo/myimage:ver` which references like a pointer to the correct images
> 
> These changes allow this job to create the `.rock` files and push them as `repo/myimage:ver-arch` (part 1)
> 
> As of now, charmhub does not support uploading a multiarch manifest -- it only supports images.
> 
> I understand if you want to reject this PR -- I'll end up building multiarch images to and pushing ghcr.io in other ways if so

Yes, there's technically no multi-arch image, but rock file is a OCI image archive file which can contain multiple manifests. 

For example, I can download all architecture of the `alpine:latest` images into one single oct image archive.

```
$ skopeo --insecure-policy copy --multi-arch all  docker://alpine:latest oci-archive:test.tar
$ 7z e -so test.tar blobs/sha256/0dfdeba6179a9c669d90362960cf052454ea8f28acbfa8338ad98c2cdb91bc42 | jq | grep architecture
        "architecture": "amd64",
        "architecture": "arm",
        "architecture": "arm",
        "architecture": "arm64",
        "architecture": "386",
        "architecture": "ppc64le",
        "architecture": "s390x",
...
```


And I can upload all architecture of the oct image archive to a local registry:

```
$ skopeo --insecure-policy copy --multi-arch all --dest-tls-verify=false oci-archive:test.tar docker://localhost:32000/test:test
Getting image list signatures
Copying 7 of 7 images in list
Copying image sha256:0deab21fa3beed803ec512c91e2a49182a7d76d91495591d73e41f479a426756 (1/7)
Getting image source signatures
Copying blob 661ff4d9561e done  
Copying config 7ffe15c5a6 done  
Writing manifest to image destination
Storing signatures
Copying image sha256:cb93cf714ba92964b24117e4b5c5b5f13b9cf72e51152376c54c209e6b34b54f (2/7)
Getting image source signatures
Copying blob 0803c38384d9 done  
Copying config 0d6528cc39 done  
Writing manifest to image destination
Storing signatures
...
```

I suppose we can achieve the same result with rockcraft packing all the architecture into the same oci iamge archive. And before upload the image to charmhub, we can just use something like `docker manifest inspect` to get the correct image sha256 for the architecture we want.

addyess:
@weiiwang01 it's cool that the oci-archive does let you store multiple images and a manifest inside it.  I wasn't aware of this

Am I wrong to assert that rockcraft doesn't currently support building on any arch other than the current running platform?  It seems that it would take an arm machine to build an arm image.  How would rockcraft resolve all these back into a single `.rock` file?

cjdcordeiro:
> @weiiwang01 it's cool that the oci-archive does let you store multiple images and a manifest inside it. I wasn't aware of this
> 
> Am I wrong to assert that rockcraft doesn't currently support building on any arch other than the current running platform? It seems that it would take an arm machine to build an arm image. How would rockcraft resolve all these back into a single `.rock` file?

Indeed, atm, rockcraft would produce a separate oci-archive per arch. It is also true that there's no support for emulation (by design) so the target arch must be compatible with the build host. 

Analogously to other craft tools, Rockcraft should, eventually, have the `remote-build` capability. Most of that feature shall rely on LP providing support for rockcraft builds. The rockcraft-specific bit of that feature implies exactly what you mentioned: being able to generate a multi-arch OCI manifest from a list of manifest, and packing all the different OCI archives into a single one. This isn't super trivial because the OCI Image spec isn't too consistent about this, i.e. the [image index](https://github.com/opencontainers/image-spec/blob/main/image-index.md) will point to an OCI manifest (`application/vnd.oci.image.manifest.v1+json`), if it is a single-arch image, **but** it will point to another index (`application/vnd.oci.image.index.v1+json) if it is a multi-arch image - so rockcraft will need to manipulate these manifest and index files with basis on the required platforms.

weiiwang01:
> > @weiiwang01 it's cool that the oci-archive does let you store multiple images and a manifest inside it. I wasn't aware of this
> > Am I wrong to assert that rockcraft doesn't currently support building on any arch other than the current running platform? It seems that it would take an arm machine to build an arm image. How would rockcraft resolve all these back into a single `.rock` file?
> 
> Indeed, atm, rockcraft would produce a separate oci-archive per arch. It is also true that there's no support for emulation (by design) so the target arch must be compatible with the build host.
> 
> Analogously to other craft tools, Rockcraft should, eventually, have the `remote-build` capability. Most of that feature shall rely on LP providing support for rockcraft builds. The rockcraft-specific bit of that feature implies exactly what you mentioned: being able to generate a multi-arch OCI manifest from a list of manifest, and packing all the different OCI archives into a single one. This isn't super trivial because the OCI Image spec isn't too consistent about this, i.e. the [image index](https://github.com/opencontainers/image-spec/blob/main/image-index.md) will point to an OCI manifest (`application/vnd.oci.image.manifest.v1+json`), if it is a single-arch image, **but** it will point to another index (`application/vnd.oci.image.index.v1+json) if it is a multi-arch image - so rockcraft will need to manipulate these manifest and index files with basis on the required platforms.

Yes, I believe we can write a script within the `operator-workflows` to merge multiple single-arch rock images into a multi-arch rock image (is a multi-architecture OCI image archive also a rock file? is there criteria an OCI image archive must meet to be qualified as a "rock" ?). And We can then use the merged multi-architecture 'rock' in all subsequent workflows. The merging while not trivial, is not overly complicated either.

cjdcordeiro:
> is a multi-architecture OCI image archive also a rock file?

yes. The rock's specification only addresses the single-arch manifest and image configuration, not the OCI's index files

addyess:
I'm closing this but let's not end the discussion of how rockcraft will build multiarch images.  I have saved this work in https://github.com/canonical/cilium-rocks/pull/4

.github/workflows/build_rocks.yaml
```diff
@@ -214,7 +248,7 @@ jobs:
       - name: Upload rock to ghcr.io
         if: steps.rock-cache.outputs.cache-hit != 'true' || inputs.cache-action == 'save'
         run: |
-          skopeo --insecure-policy copy oci-archive:$(ls "${{ matrix.path }}"/*.rock) docker://$IMAGE_REF --dest-creds "${{ github.actor }}:${{ secrets.GITHUB_TOKEN }}"
+          skopeo --insecure-policy copy oci-archive:$(ls "${{ matrix.rock.path }}"/*.rock) docker://$IMAGE_REF --dest-creds "${{ github.actor }}:${{ secrets.GITHUB_TOKEN }}"
```
yanksyoon, line 251: Hello, it seems that skopeo isn't available for ARM64, so this step also needs to be fixed.

weiiwang01: I think it should be available on ARM64 https://packages.ubuntu.com/jammy/arm64/skopeo/download

PR #250, opened 2024-02-05
Download and list any charm artifact when running the integration tests

addyess:
### Overview

Use all charm artifacts for a given pull-request in the integration test.

### Rationale

The current workflow requires that to test with a given charm artifact, there should be only one and it should be in the `charm-file` input.  That's fine for repos with a single charm, but when the repo contains multiple charms -- the tox command args should contain multiple `--charm-file` arguments.  

### Workflow Changes

the `Download charm artifact` step is changed to download all artifacts that match the `*-charm` pattern and merge into them into working-directory. (See [download-artifact](https://github.com/actions/download-artifact?tab=readme-ov-file#download-multiple-filtered-artifacts-to-the-same-directory) for details)

If the `input.charm-file` doesn't exist in the working-directory, look for all the charms downloaded  as `charm_artifacts` and add those to the args.

### Checklist

### DIFF
--- a/.github/workflows/integration_test_run.yaml
+++ b/.github/workflows/integration_test_run.yaml
@@ -247,8 +247,14 @@ jobs:
           for image in $(echo '${{ inputs.images }}' | jq -cr '.[]'); do
             args="${args} --$(echo $image | awk -F '/' '{print $NF}' | cut -d ':' -f 1)-image ${image}"
           done
+          charms_artifacts=(`find ${{ inputs.working-directory }} -maxdepth 1 -name "*.charm"`)
           if [ ! -z ${{ inputs.charm-file }} ]; then
             args="${args} --charm-file=${{ inputs.charm-file }}"
+          elif [ ${#charms_artifacts[@]} -gt 0 ]; then
+            # iterate across any local *.charm files
+            for charm in "${charms_artifacts[@]}"; do
+              args="${args} --charm-file=${charm}"
+            done
           fi
           echo "ARGS=$args" >> $GITHUB_ENV
           series=""
@@ -266,8 +272,9 @@ jobs:
         if: ${{ github.event_name == 'pull_request' }}
         continue-on-error: true
         with:
-          name: ${{ env.CHARM_NAME }}-charm
           path: ${{ inputs.working-directory }}
+          pattern: '*-charm'
+          merge-multiple: true
       - name: Run k8s integration tests
         working-directory: ${{ inputs.working-directory }}
         if: ${{ inputs.provider == 'microk8s' }}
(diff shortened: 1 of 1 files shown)


addyess:
@arturo-seijas what do you think about just downloading all the artifacts with the pattern `*-charm` and uses each one as an arg to `tox -e integration -- `.

cbartz:
I think we currently only upload one charm artifact https://github.com/canonical/operator-workflows/blob/main/.github/workflows/integration_test.yaml#L254C14-L261, so I wonder if these changes would make sense in the current state of the workflow.

addyess:
@cbartz I understand there's only one charm uploaded by this workflow.  I created a PR a while back to handle a situation where multiple charms in a repo would be built on a PR -- all with one integration test.

That PR didn't land, because there were some fatal flaws.  Much of these workflows expect a single charm artifact for a single architecture/base.  I hoped the change in this PR would be so slight that it wouldn't impact anyone who continues to use these workflows for a single charm. 

I have a repo with two charms, which I build via a matrix, then store into `k8s-charm` and `k8s-worker-charm` artifacts.  I'd like to integration test them together -- but ATM, the integration run job can't find them.

cbartz:
@addyess  Ok understood, is your repo public, can you provide a link to it please?

addyess:
> @addyess Ok understood, is your repo public, can you provide a link to it please?

@cbartz I'll give you one better -- a link to an integration test run.  Two charm-builds, one integration test run (which right  now -- rebuilds the charms because it doesn't find them locally downloaded)

https://github.com/canonical/k8s-operator/actions/runs/7792451588

.github/workflows/integration_test_run.yaml
```diff
@@ -234,6 +234,14 @@ jobs:
       - name: Pre-run script
         if: ${{ inputs.pre-run-script != '' }}
         run: bash -xe ${{ inputs.pre-run-script }}
+      - name: Download charm artifact
+        uses: actions/download-artifact@v4
+        if: ${{ github.event_name == 'pull_request' }}
```
cbartz, line 239: I think you just moved it here, but maybe you can remove that line? I think we can also run operator workflows on other events like `push`.

cbartz (dismissed):
LGTM, just a small comment.

arturo-seijas (dismissed):

.github/workflows/integration_test_run.yaml
addyess: If the artifacts are built on a push event, it should be able to download them with this pattern. If they aren't there, no pattern will match the download-artifact step should continue without failure.

cbartz: How can they be downloaded if the step is not executed with `github.event.name` set to `push`?

I have seen that you have updated the code, so I am resolving this conversation.

cbartz (approved):

arturo-seijas (approved):

PR #259, opened 2024-02-27
Allow the use of rockcraft pulled and built from source to build rock images.

weiiwang01:
Add the `rockcraft-repository` and `rockcraft-ref` options to the integration workflow to pull and build rockcraft from that specific source, and use the rockcraft built from source to build rocks in the integration tests. This is to enable the development of some experimental features in rockcraft that are not available in the snap store version of rockcraft.

Also, remove the rockcraft cache system, as the adoption has not been very great, and considering the risk of creating stalled rock images if the cache is enabled. Perhaps we will have an improved cache system in the future.

### DIFF
--- a/.github/workflows/build_rocks.yaml
+++ b/.github/workflows/build_rocks.yaml
@@ -25,10 +25,6 @@ on:
         type: string
         description: The working directory for jobs
         default: "./"
-      cache-action:
-        type: string
-        description: The cache action can either be "save" or "restore".
-        default: restore
       rockcraft-ref:
         description: Used in conjunction with rockcraft_repo to pull and build rockcraft from source instead of using snapstore version.
         type: string
@@ -131,11 +127,6 @@ jobs:
         with:
           path: ./rockcraft*.snap
           key: ${{ steps.restore-rockcraft.outputs.cache-primary-key }}
-      - uses: actions/upload-artifact@v4
-        if: inputs.rockcraft-repository != ''
-        with:
-          name: rockcraft-snap
-          path: "rockcraft*.snap"
       - name: Install Rockcraft
         if: inputs.rockcraft-repository != ''
         run: sudo snap install --dangerous --classic rockcraft*.snap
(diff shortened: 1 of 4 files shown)


arturo-seijas (dismissed):
Can you provide a successful run?

weiiwang01:
> Can you provide a successful run?

I think the test failure is caused by the container cache, let me rerun this.
But for the new feature, I have already used it in https://github.com/canonical/xiilib/actions/runs/8075172173/job/22062013902?pr=4

arturo-seijas:
> > Can you provide a successful run?
> 
> I think the test failure is caused by the container cache, let me rerun this. But for the new feature, I have already used it in https://github.com/canonical/xiilib/actions/runs/8075172173/job/22062013902?pr=4

Cool, thanks!

javierdelapuente (dismissed):

javierdelapuente (dismissed):

weiiwang01:
@javierdelapuente thanks for helping me fixing the rockcraft packing issue!

.github/workflows/build_rocks.yaml
```diff
@@ -25,10 +25,14 @@ on:
         type: string
         description: The working directory for jobs
         default: "./"
-      cache-action:
+      rockcraft-ref:
+        description: Used in conjunction with rockcraft_repo to pull and build rockcraft from source instead of using snapstore version.
```
yanksyoon, line 29:
```suggestion
description: Used in conjunction with rockcraft-repository to pull and build rockcraft from source instead of using snapstore version.
```
just using the full input var name so it's less confusing :)

.github/workflows/build_rocks.yaml
```diff
@@ -88,6 +89,47 @@ jobs:
...
     steps:
+      - uses: actions/checkout@v4
+        if: inputs.rockcraft-repository != ''
+        with:
+          repository: ${{ inputs.rockcraft-repository }}
+          ref: ${{ inputs.rockcraft-ref }}
+          path: rockcraft
+          fetch-depth: 0
+      - name: Get Rockcraft SHA
+        if: inputs.rockcraft-repository != ''
```
yanksyoon, line 100: Should the inputs.... be within ${{ }}?

yanksyoon: Same with other inputs :D

yanksyoon (dismissed):
LGTM, minor comments :)

.github/workflows/build_rocks.yaml
weiiwang01: Updated, thanks!

.github/workflows/build_rocks.yaml
weiiwang01: The `${{ }}` can be omitted in the `if` statement: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif

javierdelapuente (approved):

yanksyoon (approved):

PR #268, opened 2024-03-26
Specify charm directory to allow paas-app-charmer apps

javierdelapuente:
Applicable spec: ISD124

### Overview

paas-app-charmer projects have a different project structure. The `charmcraft.yaml` file is not in the `working-directory`. For `paas-app-charmer` apps, the `rockcraft.yaml` file is in the `working-directory` and the `charmcraft.yaml` file inside a directory (the new `charm-working-directory` variable). Besides, the docs directory is inside the `charm-working-directory` (`discourse-gatekeeper` has also to be informed where the charm directory is).

The new `charm-working-directory` input variable is included in several workflows, and specifies the working directory for the charm. This input variable supersedes `docs-working-directory`, so this PR is not backward compatible with projects using that variable (that input variable was added for paas-app-charmer, and it does not make sense now, as the docs directory is by convention under the charm directory)

In the publish_charm workflow, the new input `paas-app-charmer-oci-name` has been added, as the resource oci image on paas-app-charmer is not the same as the rockcraft name plus "-image".

### Rationale

paas-app-charmer projects will have a subdirectory with the charm and under this one the docs. That information should be given to discourse-gatekeeper for the documentation and to charmcraft when necessary.

### Workflow Changes

Added new input variable `charm-working-directory` in workflows where the charm directory is needed.
Added new input variable `paas-app-charmer-oci-name` in publish_charm workflow to correct the name of the oci resource.

### Checklist


yhaliaw (dismissed):

yanksyoon (dismissed):

.github/workflows/promote_charm.yaml
```diff
@@ -10,16 +10,16 @@ on:
...
+        type: string
+        description: The working directory for the charm
       destination-channel:
         type: string
         description: 'Destination Channel'
       doc-automation-disabled:
         type: boolean
         description: 'Whether to disable the documentation automation'
         default: true
-      docs-working-directory:
```
arturo-seijas, line 20: This actually prevents the docs to be in a place different from the charm or am I missing something?

arturo-seijas (commented):
Can you update the readme to match these changes?

arturo-seijas (dismissed):

.github/workflows/publish_charm.yaml
```diff
@@ -19,12 +22,22 @@ on:
...
   REGISTRY: ghcr.io
   OWNER: ${{ github.repository_owner }}
 
 jobs:
+  vars:
+    runs-on: ubuntu-22.04
+    outputs:
+      charm_working_directory: ${{ inputs.charm-working-directory || inputs.working-directory }}
+    steps:
+      - run: echo "Exposing env vars"
```
cbartz, line 40: Why not add `charm_working_directory` as a workflow env (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#env) so its available to all jobs without the need for a new job (requesting a runner which counts against our organisational quota)?

.github/workflows/publish_charm.yaml
```diff
@@ -231,7 +250,14 @@ jobs:
...
           
           for image in "${images[@]}"; do
-            resource=$(echo $image | awk -F '/' '{print $NF}' | cut -d ':' -f 1)-image
+            resource_name=$(echo $image | awk -F '/' '{print $NF}' | cut -d ':' -f 1)
+            # This is for paas-app-charmer for Django, as the main image name is named "django-app-image" and
+            # should match the main rockcraft.yaml.
+            if [[ ${resource_name} == ${CHARM_NAME} ]] && [[ " ${resources[@]} " =~ " django-app-image " ]]; then
+                resource_name=django-app
+            fi
+            resource=${resource_name}-image
```
cbartz, line 259: Could we then also add the mechanism for flask? I guess it would be `flask-app-image` and `flask-app`. Is this correct @weiiwang01?

.github/workflows/publish_charm.yaml
javierdelapuente: mmm I think I read that context cannot be used in defaults.run (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaultsrun). Let me try...

javierdelapuente: https://github.com/canonical/operator-workflows/actions/runs/8571566147

javierdelapuente: I will try in the step.working-directory

javierdelapuente: Also for `uses`.... https://github.com/canonical/operator-workflows/actions/runs/8572352463 I will go to the original solution with outputs

cbartz: Ok, thanks for trying. Its a bit strange, as https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability indicates that its possible to use env in `defaults.run`, although https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaultsrun actually states that its not possible to use contexts.

.github/workflows/publish_charm.yaml
weiiwang01: Yes, I think it would be good to do the same for Flask, as it has the same issue where the `name` in `rockcraft.yaml` doesn't match `resources` in `charmcraft.yaml` or `metadata.yaml`. Thanks!

.github/workflows/integration_test_run.yaml
```diff
@@ -250,8 +253,9 @@ jobs:
           path: ${{ inputs.working-directory }}
           pattern: '*-charm'
           merge-multiple: true
+
       - name: Integration tests variable setting
-        working-directory: ${{ inputs.working-directory }}
+        working-directory: ${{ inputs.charm-working-directory || inputs.working-directory }}
```
weiiwang01, line 258: I was thinking that perhaps the `charm-working-directory` could be expressed as `charm-directory`, which would represent the relevant path to the charm directory from the working directory. The default value would be `charm-directory : "."`, and for charms located in `./charm/`, it would be `charm-directory: "charm"`. With this, the final charm working directory would be `{{ inputs.working-directory }}/{{ inputs.charm-directory }}`.

However, I believe the current solution is perfectly fine as well, so it's up to you.

weiiwang01 (dismissed):

arturo-seijas (dismissed):

arturo-seijas (approved):

weiiwang01 (approved):

addyess:
@javierdelapuente 

oops, these weren't the only workflows using `docs-working-directory`

PR #285, opened 2024-04-22
Better build and publish pipeline

weiiwang01:
Rewrite the build and publish workflows by using TypeScript and GitHub Action toolkits, replacing the majority of the existing Bash script logic. This enhancement not only modernizes the workflow but also introduces several improvements:

- The integration test workflow now support charms and rocks with multiple bases.
- Integration and publish workflows for pull requests originating from forked repositories will automatically use action artifacts, instead of registry which causes permission issues.
- The integration test workflow supports concurrent job execution without artifact conflicts.
- Small enhancement in the parallelization and acceleration of the image build process.
- The operator-workflows itself include more efficient parallelized self-checks.
- Users can now use the publish workflow with an ad-hoc integration test run outside of pull requests.


cbartz (commented):
Reviewed until README.md

.github/workflows/publish_charm.yaml
```diff
@@ -33,36 +34,25 @@ on:
...
+          If these are not provided, the system will default to using charms and rocks from the most 
+          recent successful integration test that matches the git tree ID.
+        type: string
+        default: ""
 env:
   REGISTRY: ghcr.io
   OWNER: ${{ github.repository_owner }}
 
 jobs:
-  branch-up-to-date-check-enabled:
```
arturo-seijas, line 42: I don't see this as part of the action

.github/workflows/integration_test.yaml
```diff
@@ -160,199 +175,135 @@ on:
...
-          fi
-          echo "IMAGES=$IMAGES" >> $GITHUB_ENV
-  build-charm:
-    name: Build and push charm
-    needs: get-runner-image
-    runs-on: ${{ needs.get-runner-image.outputs.runs-on }}
-    outputs:
-      charm-file: ${{ env.CHARM_FILE }}
+      - uses: actions/checkout@v4.1.1
+      - uses: canonical/operator-workflows/internal/plan@better-build
```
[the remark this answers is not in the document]
weiiwang01: Indeed, but setting this to `@main` in the pull request would cause the tests to fail, as the actions do not yet exist in the main branch at this point. Therefore, I will need to update all the references in the final step of this pull request, just before merging.

tests/workflows/integration/test-upload-charm/tox.ini
```diff
@@ -118,7 +118,7 @@ commands =
 description = Run integration tests
 deps =
     juju==3.1.2.0
-    pytest
+    pytest==8.1.1
```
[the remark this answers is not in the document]
weiiwang01: Removed, thanks!

tests/workflows/integration/test-upload-charm/tox.ini
```diff
@@ -105,7 +105,7 @@ commands =
 description = Run integration tests
 deps =
     juju==2.9.42.4
-    pytest
+    pytest==8.1.1
```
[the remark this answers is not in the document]
weiiwang01: Removed, thanks!

.github/workflows/publish_charm.yaml
```diff
@@ -77,23 +67,17 @@ jobs:
...
-    name: Get runner image
-    uses: ./.github/workflows/get_runner_image.yaml
-    with:
-      working-directory: ${{ inputs.working-directory }}/${{ inputs.charm-directory }}
+          charm_dir: ${{ needs.publish-charm.outputs.charm-directory }}
   release-charm-libs:
     name: Release charm libs
-    needs: [branch-up-to-date-check-enabled]
     runs-on: ubuntu-22.04
+    needs: [ publish-charm ]
```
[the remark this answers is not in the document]
weiiwang01: Updated, thanks!

.github/workflows/publish_charm.yaml
```diff
@@ -33,36 +34,25 @@ on:
...
-      - run: |
-          if [ ${{ fromJson(steps.get-branch-protection.outputs.data).strict }} != "true" ]; then
-            echo "::error::Strict checks are not enabled for this repository"
-            exit 1
-          fi
   draft-publish-docs:
     name: Draft publish docs
-    needs: [branch-up-to-date-check-enabled]
     runs-on: ubuntu-22.04
+    needs: [publish-charm]
```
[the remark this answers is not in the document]
weiiwang01: Updated, thanks!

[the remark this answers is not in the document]
weiiwang01: Updated, thanks!

.github/workflows/publish_charm.yaml
arturo-seijas: The intent of this step was to prevent pushing artifacts that might not contain all the changes. I'll illustrate with an example, it's easier:
- We have to open PRs with successful CI checks
- PR A is merged and the artifacts are retrieved from the build and pushed to charmhub. -> OK
- PR B is merged and the artifacts are retrieved from the build and pushed to charmhub. If the check is disabled, the code does not have to be rebased, hence the arctifacts do not contain feature A. If the check is enabled, the PR will have to be rebased, triggering a new build from which the artifacts will be fetched

weiiwang01: Okay, I understand what you're saying. This issue is resolved because the publish workflow only retrieves artifacts that share the same git tree id. Since the file contents is always the same when two commits have the same git tree id, this ensures that the artifact is constructed from identical source code. In the case of non-linear history repositories, the publish workflow will just fail.

arturo-seijas: Understood

weiiwang01: I update the error message to recommend enable branch update check if it can't find the matching tree id, hope this can help

https://github.com/canonical/operator-workflows/blob/679bfe7624df6957adc37bc3b7b326fc02f3032c/src/publish.ts#L79-L82

.github/workflows/integration_test.yaml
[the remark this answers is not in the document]
cbartz: Yes, that is why I meant **after merging** .

javierdelapuente (commented):
would it be useful to try to create some unit tests for ts files/functions/classes?

weiiwang01:
> would it be useful to try to create some unit tests for ts files/functions/classes?

Yes, the problem is that a lot of the functions require to access Github API or git repository when running, so it might be a little difficult to add unit tests for them. I have added more integration tests in this pull request, and I will definitely try to add more unit tests in the future.

src/publish.ts
```diff
@@ -0,0 +1,323 @@
...
+    this.octokit = github.getOctokit(this.token)
+    this.workflowFile = core.getInput('workflow-file')
+    this.workingDir = core.getInput('working-directory')
+    this.resourceMapping = JSON.parse(core.getInput('resource-mapping'))
+    this.workflowRunId = core.getInput('workflow-run-id')
+    this.identifier = core.getInput('identifier')
+    this.artifact = new DefaultArtifactClient()
+  }
+
+  async findWorkflowRunId(): Promise<number> {
```
cbartz, line 35: You may want to use step-down rules and ordering features to improve readability: https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#function-and-method-ordering

tests/workflows/integration/test-rock/requirements.txt
```diff
@@ -0,0 +1 @@
+ops
```
cbartz, line 1: final new line

package.json
```diff
@@ -0,0 +1,72 @@
+{
+  "name": "operator-workflows",
+  "description": "Github workflows for charm developments",
+  "version": "0.0.0",
```
yanksyoon, line 4:
```suggestion
  "version": "0.0.1",
  ```

File: package.json [hunk 10]
```diff
@@ -0,0 +1,72 @@
...
+    "actions",
+    "node",
+    "setup"
+  ],
+  "exports": {
+    ".": "./dist/index.js"
+  },
+  "engines": {
+    "node": ">=20"
+  },
```
yanksyoon, line 25: I think the .node-version file above might not be necessary if i understand correctly? :D

src/build.ts
```diff
@@ -0,0 +1,372 @@
...
+      { name: params.plan.name, files: charmFiles.map(f => path.basename(f)) },
+      null,
+      2
+    )
+  )
+  await artifact.uploadArtifact(
+    params.plan.output,
+    [...charmFiles, manifestFile],
+    params.plan.source_directory
+  )
```
yanksyoon, line 125: I'm wondering if overwrite:true is required for uploadArtifact here?

yanksyoon: https://github.com/actions/upload-artifact/?tab=readme-ov-file#overwriting-an-artifact

yanksyoon: And perhaps for other uploadArtifact calls below if this is valid

src/build.ts
```diff
@@ -0,0 +1,372 @@
...
+      'charmcraft',
+      '--channel',
+      params.charmcraftChannel,
+      '--classic'
+    ])
+  } else {
+    await exec.exec('sudo', ['snap', 'install', 'charmcraft', '--classic'])
+  }
+  core.startGroup('charmcraft pack')
+  await exec.exec('charmcraft', ['pack', '--verbosity', 'trace'], {
```
cbartz, line 104: Is the EXPERIMENTAL flag required for paas-app charmer (e.g. django)?

src/plan.ts
```diff
@@ -0,0 +1,186 @@
...
+    ...(await planBuildDockerImage(workingDir, id, imageOutputType))
+  ]
+}
+
+export async function run(): Promise<void> {
+  try {
+    let id = `${new Date().toISOString().replaceAll(':', '-').replace(/\..+/, '')}-${crypto.randomUUID().split('-')[3]}`
+    const identity = core.getInput('identifier')
+    if (identity.includes('__')) {
+      core.setFailed('identifier can not contain "__"')
```
yanksyoon, line 138: Out of curiosity, may I know the reason behind this?

src/plan.ts
```diff
@@ -0,0 +1,186 @@
...
+      default:
+        core.setFailed(`unknown upload-image input: ${uploadImage}`)
+        return
+    }
+    const buildPlans = await planBuild(workingDir, id, imageOutputType)
+    const plan: Plan = {
+      working_directory: workingDir,
+      build: buildPlans
+    }
+    core.info(`generate workflow plan: ${JSON.stringify(plan, null, 2)}`)
```
yanksyoon, line 166:
```suggestion
    core.info(`generated workflow plan: ${JSON.stringify(plan, null, 2)}`)
```

src/plan.ts
```diff
@@ -0,0 +1,186 @@
...
+
+async function planBuild(
+  workingDir: string,
+  id: string,
+  imageOutputType: 'file' | 'registry'
+): Promise<BuildPlan[]> {
+  return [
+    ...(await planBuildCharm(workingDir, id)),
+    ...(await planBuildRock(workingDir, id, imageOutputType)),
+    ...(await planBuildDockerImage(workingDir, id, imageOutputType))
```
yanksyoon, line 129: I'm wondering if the planBuild should be an object 
```typescript
{charmBuildPlan: BuildPlan[], rockBuildPlan: BuildPlan[], dockerBuildPlan: BuildPlan[]}
``` 
rather than a list - the reason is because we cannot differentiate different build plans from each other from the output - even though we assume there will be one for each.

tests/workflows/integration/test-rock/requirements.txt
yanksyoon, line 1: Minor missing newline :D

src/publish.ts
```diff
@@ -0,0 +1,323 @@
...
+      core.endGroup()
+      core.startGroup('retrieve charm info')
+      const {
+        name: charmName,
+        dir: charmDir,
+        files: charms
+      } = await this.getCharms()
+      core.endGroup()
+      core.info(
+        `start uploading image resources: ${JSON.stringify(Object.fromEntries([...images]))}`
```
yanksyoon, line 98: Out of curiosity, is there a reason why you're spreading the images?

yanksyoon (dismissed):
Wowzas! LGTM, a few questions (first round of reviews done, I think i'll need to read again after another coffee)

.github/workflows/integration_test.yaml
```diff
@@ -160,199 +175,135 @@ on:
...
-        if: steps.restore-charmcraft.outputs.cache-hit != 'true' && inputs.charmcraft-repository != ''
+  scan:
+    name: Scan Image (${{ matrix.scan.file }})
+    runs-on: ubuntu-latest
+    needs: [ plan-scan ]
+    strategy:
+      matrix:
+        scan: ${{ fromJSON(needs.plan-scan.outputs.scans) }}
+    steps:
+      - uses: actions/download-artifact@v4
```
arturo-seijas, line 240: Can you pin this to 4.1.7?

.github/workflows/integration_test.yaml
```diff
@@ -160,199 +175,135 @@ on:
...
-          echo "CHARM_FILE=$(ls $CHARM_NAME_*.charm || echo UNKNOWN)" >> $GITHUB_ENV
-
-      - name: Upload charm artifact
-        if: ${{ !contains(fromJson('["", "UNKNOWN"]'), env.CHARM_FILE) && !cancelled() }}
-        uses: actions/upload-artifact@v4
+          docker image pull ${{ matrix.scan.image }}
+          docker image save -o ${{ matrix.scan.file }} ${{ matrix.scan.image }}
+      - name: Run Github Trivy Image Action
+        uses: aquasecurity/trivy-action@0.19.0
+        if: steps.rock-cache.outputs.cache-hit != 'true' || inputs.cache-action == 'save'
```
arturo-seijas, line 254: Where is the step `rock-cache` defined?

.licenserc.yaml
```diff
@@ -25,4 +25,11 @@ header:
     - '**/icon.svg'
     - '**/LICENSE'
     - '**/trivy.yaml'
+    - '.eslintignore'
+    - '.gitattributes'
+    - '.node-version'
+    - '.prettierignore'
+    - 'dist/**'
+    - 'internal/**'
+    - 'src/**'
```
arturo-seijas, line 34: src? Is this intended?

package.json
```diff
@@ -0,0 +1,72 @@
...
+    "package:plan": "npx ncc build src/plan.ts -o dist/plan --source-map --license licenses.txt",
+    "package:plan-scan": "npx ncc build src/plan-scan.ts -o dist/plan-scan --source-map --license licenses.txt",
+    "package:plan-integration": "npx ncc build src/plan-integration.ts -o dist/plan-integration --source-map --license licenses.txt",
+    "package:publish": "npx ncc build src/publish.ts -o dist/publish --source-map --license licenses.txt",
+    "test": "npx jest",
+    "all": "npm run format:write && npm run lint && npm run test && npm run coverage && npm run package"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@actions/artifact": "^2.1.5",
```
arturo-seijas, line 44: Can we pin these to an exact version?

src/build.ts
weiiwang01: The new `operator-workflows` will ensure each job uploads artifacts with different names, so overwrite is not required here, thanks!

src/plan.ts
weiiwang01: Rock, charm and docker-image are actually built by the same `build` job, so at here all types of plans are put into a list. This is to simplify the Github action run matrix setup. Thanks!

src/plan.ts
weiiwang01: Because there are many restrictions in artifact names, so `__` is chosen as the separator in the artifact names. Restrict the use of `__` in user provided input to avoid ambiguity.

src/plan.ts
weiiwang01: Updated, thanks!

tests/workflows/integration/test-rock/requirements.txt
weiiwang01: updated, thanks!

weiiwang01: updated, thanks!

package.json
weiiwang01: Updated, thanks!

src/publish.ts
```diff
@@ -0,0 +1,323 @@
...
+      )
+    }
+    const runIter = this.octokit.paginate.iterator(
+      this.octokit.rest.actions.listWorkflowRuns,
+      {
+        owner,
+        repo,
+        workflow_id: workflowResp.data.id,
+        status: 'success',
+        created: `>${new Date(Date.now() - 7 * 24 * 60 * 60 * 1000).toISOString().slice(0, 10)}`
```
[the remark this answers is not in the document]
weiiwang01: Typically, the publish workflow is triggered immediately following the merge of a pull request. The time limit is just a precaution to prevent the upload of outdated charm artifacts in case the publish workflow not executed for a long time after the integration test.

src/publish.ts
weiiwang01: I am not very sure what spreading the images means here.

dist/build/licenses.txt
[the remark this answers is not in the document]
weiiwang01: Yes, I misused a command line argument in the compile pipeline, removed, thanks!

.licenserc.yaml
weiiwang01: Removed, thanks!

.github/workflows/integration_test.yaml
```diff
@@ -160,199 +175,134 @@ on:
...
-        if: steps.restore-charmcraft.outputs.cache-hit != 'true' && inputs.charmcraft-repository != ''
+  scan:
+    name: Scan Image (${{ matrix.scan.file }})
+    runs-on: ubuntu-latest
+    needs: [ plan-scan ]
+    strategy:
+      matrix:
+        scan: ${{ fromJSON(needs.plan-scan.outputs.scans) }}
+    steps:
+      - uses: actions/download-artifact@v4.17
```
arturo-seijas, line 240: I believe you missed a dot here. The last version is 4.1.7

weiiwang01: Updated, thanks!

arturo-seijas (approved):

src/publish.ts
yanksyoon: Oh sorry i should have clarified, i was referring to the use of ... spread operator inside a list [...images]. Would it be redundant here?

yanksyoon (approved):
Thanks for the clarifications, LGTM!

src/publish.ts
weiiwang01: Yes, I think `...images` doesn't create an iterator like Python so I think the `[]` is necessary here.

package.json
weiiwang01: I think `.node-version` might still be necessary for tools like `nodenv` or `nvm`. It was copied from https://github.com/actions/typescript-action I think we can keep it here for now.

addyess:
@weiiwang01 @arturo-seijas @yanksyoon @cbartz 

I've got broken publishes now :(
https://github.com/canonical/k8s-operator/actions/runs/8970206205/job/24633223435

weiiwang01:
> @weiiwang01 @arturo-seijas @yanksyoon @cbartz
> 
> I've got broken publishes now :( https://github.com/canonical/k8s-operator/actions/runs/8970206205/job/24633223435

Let me see

addyess:
> > @weiiwang01 @arturo-seijas @yanksyoon @cbartz
> > I've got broken publishes now :( https://github.com/canonical/k8s-operator/actions/runs/8970206205/job/24633223435
> 
> Let me see

there's no `metadata.yaml` in my charm.  It only uses `charmcraft.yaml`

weiiwang01:
> > > @weiiwang01 @arturo-seijas @yanksyoon @cbartz
> > > I've got broken publishes now :( https://github.com/canonical/k8s-operator/actions/runs/8970206205/job/24633223435
> > 
> > 
> > Let me see
> 
> there's no `metadata.yaml` in my charm. It only uses `charmcraft.yaml`

Yes, the publish workflow doesn't consider the case of charm doesn't have any resources, let me update the action.

addyess:
@weiiwang01 


why do the workflows point to a different branch as well?

```
      - uses: canonical/operator-workflows/internal/publish@better-build
```

https://github.com/search?q=repo%3Acanonical%2Foperator-workflows%20%20better-build&type=code

yanksyoon:
@addyess This was to make the tests on this repo pass - I believe weii was already on a followup PR to address this

weiiwang01:
> @weiiwang01
> 
> why do the workflows point to a different branch as well?
> 
> ```
>       - uses: canonical/operator-workflows/internal/publish@better-build
> ```
> 
> https://github.com/search?q=repo%3Acanonical%2Foperator-workflows%20%20better-build&type=code

This is temporary because the reusable workflow and GitHub workflow actions use different reference addressing; it will be changed to `@main` soon. I have fixed the `metadata.yaml` issue. Can you try rerunning the workflow?

src/publish.ts
yanksyoon: From what I understand, obect.fromEntries creates an object from list of key-value pair (list), would images already be in such format?

addyess:
@weiiwang01 
https://github.com/canonical/k8s-operator/actions/runs/8970206205

weiiwang01:
> @weiiwang01 https://github.com/canonical/k8s-operator/actions/runs/8970206205

Sure, here's a corrected version:

Okay, I believe the integration test run for the corresponding pull request https://github.com/canonical/k8s-operator/pull/97 is using the old version of the workflow, while the publish workflow is using the newer version that uses a different artifact schema. Could you please rerun the integration tests in that pull request and then rerun the publish workflow?

addyess:
@weiiwang01 whee!  none of this fun while trying to make a release.  When this gets solid can i request a `v0` tag?

addyess:
@weiiwang01 which piece of the "old version of the integration tests workflow" is necessary for the new publish to work?

cbartz:
@addyess  If its urgent, pin your workflows to use operator-workflows with the branch: `before-better-build` . Have just pushed it.

So e.g.
```
uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@before-better-build
```

Issue #290, opened 2024-05-01
Publishing charms from release branches check the branch-protections of the main branch


addyess:
The publish-charm workflow only ever checks the `main` branch for the branch protection rules, and ignores release branches.  The workflow should at least accept an alternate branch or use the branch based on the workflow event which is triggering this job:

```yaml
  branch-up-to-date-check-enabled:
    runs-on: ubuntu-22.04
    steps:
      - uses: octokit/request-action@v2.3.0
        id: get-branch-protection
        env:
          GITHUB_TOKEN: ${{ secrets.REPO_ACCESS_TOKEN }}
        with:
          route: GET /repos/{owner}/{repo}/branches/{branch}/protection/required_status_checks
          repo: ${{ github.event.repository.name }}
          owner: ${{ github.event.repository.owner.login }}
          branch: main       # <---------- fixed to the main branch
      - run: |
          if [ ${{ fromJson(steps.get-branch-protection.outputs.data).strict }} != "true" ]; then
            echo "::error::Strict checks are not enabled for this repository"
            exit 1
          fi
```

cbartz:
I think we no longer check for branch protection rules since a major rewrite, correct @weiiwang01 ?

weiiwang01:
> I think we no longer check for branch protection rules since a major rewrite, correct @weiiwang01 ?

Yes, I think they have all been removed.

PR #291, opened 2024-05-01
Check branch-protection of branch where this workflow is run

addyess:
Issue: https://github.com/canonical/operator-workflows/issues/290

### Overview

Remove the hard-coded link to the `main` branch on the charm to support branch protection checks on release branches

### Rationale

On push to a main branch, it makes sense to push to `latest/edge`, so one should check the branch protection rules of `main`
On push to a `release-xxx` branch, a charm could publish the xxx/edge and ought to check branch protection rules of that `release-xxx` branch

### Workflow Changes

Look up the active branch based on the workflow event:

### Checklist

### DIFF
--- a/.github/workflows/publish_charm.yaml
+++ b/.github/workflows/publish_charm.yaml
@@ -42,6 +42,11 @@ jobs:
   branch-up-to-date-check-enabled:
     runs-on: ubuntu-22.04
     steps:
+      - env:
+          BRANCH: ${{ github.base_ref || github.ref }}
+        run: |
+          BRANCH=${BRANCH#refs/heads/}  # strip off refs/heads/ if it exists
+          echo "BRANCH=${BRANCH}" >> "$GITHUB_OUTPUT"
       - uses: octokit/request-action@v2.3.0
         id: get-branch-protection
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_charm.yaml
```diff
@@ -42,6 +42,11 @@ jobs:
   branch-up-to-date-check-enabled:
     runs-on: ubuntu-22.04
     steps:
+      - env:
+          BRANCH: ${{ github.base_ref || github.ref }}
```
cbartz, line 46: I don't think this will work if the workflow is triggered by the `release` event, as `github_ref` will be equal to a tag (which cannot have branch protection with status check). We should add a short explanation to the docs with the supported events.
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

cbartz (commented):
@addyess You may be interested in reviewing https://github.com/canonical/operator-workflows/pull/285 , which will also change the Publish Charm workflow.

.github/workflows/publish_charm.yaml
cbartz: Yes, but my point is that other users might try to trigger the workflow on the `release` event, which might fail. Maybe an edge case, but I still think it is worth adding a one-liner in the docs about which workflow events are supported (push, workflow_dispatch, pull_request) to avoid having to troubleshoot this in the future if someone complains.

cbartz:
/canonical/self-hosted-runners/run-workflows d158e2224d77eff10ed591b5ee084e25e55594e3

.github/workflows/publish_charm.yaml
arturo-seijas: @addyess , if you could add a small explanation with this in the readme that'd be great

PR #293, opened 2024-05-02
Support publishing artifacts from a GH run with different artifact names

addyess:
Charm and Image gh artifacts are generally stored with the following names
* ${CHARM_NAME}-charm
* ${CHARM_NAME}-images

If publishing for multiple architectures, it may be necessary for the same action run to publish multiple artifacts with different names for the same charm.

See [this PR run](https://github.com/canonical/k8s-operator/actions/runs/8928656974/job/24524990044) 
where the same charm creates multiple charm artifacts based on the support architectures.

When the publish workflow is called, it should pull the appropriate artifacts from the run.  
The example PR run above has artifacts named `k8s-arm64-charm` and `k8s-amd64-charm` produced.  For the call to this workflow to complete, i will update the call to include the `artifact-format` input:

```yaml
    with:
      channel: latest/edge
      working-directory: ${{ matrix.charm.path }}
      tag-prefix: ${{ matrix.charm.tagPrefix }}
      artifact-format: '${CHARM_NAME}-${{matrix.strategy.arch}}'
```

### DIFF
--- a/.github/workflows/publish_charm.yaml
+++ b/.github/workflows/publish_charm.yaml
@@ -10,6 +10,20 @@ on:
         required: true
         type: string
         description: 'Destination Channel'
+      artifact-format:
+        type: string
+        default: ""
+        description: |
+          Charm and Image gh artifacts are generally stored with the following names
+            * ${CHARM_NAME}-charm
+            * ${CHARM_NAME}-images
+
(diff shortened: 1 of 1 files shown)


addyess:
@cbartz and @arturo-seijas Hope you'll consider this or offer a better suggestion

PR #313, opened 2024-05-09
build on configurable architecture

addyess:
### Overview
Support multi-arch builds of charms and images

### Details
In light of the new better-builder which plans builds and integration tests, I'd like to propose we allow the workflows to be called from multiple architectures. 

I have charms which need to be built for ARM64 and AMD64... and while they could be planned together, the better-builder does not yet support that.  Like-wise, to run the integration tests on the resultant charms, the integration tests require different architectures as well.  This proposal introduces a `builder-runner-label` which defaults to `ubuntu-latest` but can also be set to `ARM64` to engage the self-hosted runners in the organization to build the arm64 version of the charms.  Then using the `self-hosted-runner-label`, I can have integration tests ALSO run on each architecture.

Lastly, using the new `identifier` field, i can identify the plans with either an `amd64` or `arm64` architecture. 

Replaces work done for: https://github.com/canonical/operator-workflows/pull/293


addyess:
@weiiwang01 @arturo-seijas @cbartz 

I hope this works is clear for enabling arm64 builds of charms.  I do like the build plan feature and i've tried to incorporate it here:

https://github.com/canonical/k8s-operator/pull/94

arturo-seijas (approved):
LGTM, thx

amandahla (approved):

arturo-seijas:
/canonical/self-hosted-runners/run-workflows 23599d022a3489bad04054b75156e227913af54c

addyess:
> @arturo-seijas why did this get closed? Was there something i should have done better?

oh, you've re-opened.  Disregard 👍🏻

PR #318, opened 2024-05-15
default architecture of self-hosted runner to x64, allowing for others

### LINKED ISSUE #317: Running integration from ARM on self-hosted runners
https://github.com/canonical/operator-workflows/blob/main/.github/workflows/integration_test_run.yaml#L161-L167

Is really nice to be able to `run-on` any runner you wish or -- if self-hosted, add an extra tag to the runner

but what breaks it is this:

```yaml
      self-hosted-runner: true
      self-hosted-runner-label: ARM64
```

primarily because the labels on the runner become this:

```
Requested labels: self-hosted, x64, jammy, ARM64
```

oops -- that's not what we want. 

`x64` and `ARM64` labels on the same worker probably isn't going to work out

addyess:
Resolves #317 by defaulting self-hosted runners to x64 but allowing for changes to ARM runners.

Since there are other architectures available to self-hosted runners, we should allow the workflows to target architecture labels other than just x64.  With a new input passed to the workflows, `self-hosted-runner-arch` defaulted to `x64`, downstream charm workflows can run on other architectures such as ARM.

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -93,6 +93,10 @@ on:
         type: boolean
         description: Whether to use self-hosted runners to run the jobs.
         default: false
+      self-hosted-runner-arch:
+        type: string
+        description: Architecture to use on self-hosted runners to run the jobs.
+        default: "x64"
       self-hosted-runner-label:
         type: string
         description: Label for selecting the self-hosted runners.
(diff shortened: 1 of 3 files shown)


addyess:
thanks @arturo-seijas and @weiiwang01 for your review

arturo-seijas:
/canonical/self-hosted-runners/run-workflows 81cd09b3a06b3404a74a14501d444530fb4cd8c9

arturo-seijas (approved):

codethulu (approved):

PR #336, opened 2024-06-03
add_charmcraft_channel_input

Thanhphan1147:
The charmcraft channel used in https://github.com/canonical/charming-actions can take the charmcraft channel as input, we need to support this as the yaml schema for `charmcraft.yaml` have significant differences between charmcraft v2 and v3 and a v3 schema will not pass validation with charmcraft v2

### Checklist

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -10,6 +10,10 @@ on:
         type: string
         description: The working directory for the docs
         default: "."
+      charmcraft-channel:
+        description: Charmcraft channel to use
+        type: string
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -10,6 +10,10 @@ on:
         type: string
         description: The working directory for the docs
         default: "."
+      charmcraft-channel:
```
arturo-seijas, line 13: There should be a place where charmcraft is installed. We should change that one too for consistency

PR #345, opened 2024-06-10
Add support for custom ubuntu series when selecting runners

Thanhphan1147:
Since we are planning to support custom image for self-hosted runners, we should remove the hard-coded `jammy` label. This PR proposes to replace it with a new input argument `self-hosted-runner-image`

### Checklist


.github/workflows/test.yaml
```diff
@@ -29,6 +29,10 @@ on:
         type: string
         description: Label for selecting the self-hosted runners.
         default: "large"
+      self-hosted-runner-series:
```
cbartz, line 32:
```suggestion
      self-hosted-runner-image:
```
We are moving into custom image support for the self-hosted runners, so there may be different types of images supported. We also use the term `image` more than series in this context.

.github/workflows/test.yaml
```diff
@@ -29,6 +29,10 @@ on:
...
+        type: string
+        description: Flavor of the requested runner. Supports only 'jammy' or 'noble'.
```
cbartz, line 34:
```suggestion
        description: Image of the requested runner. Supports only 'jammy' or 'noble'.
```

.github/workflows/test.yaml
Thanhphan1147: I see, updating the input name, thanks!

Thanhphan1147:
I think we can roll-out this change first on only the `test` workflow, and port it to other workflows once it has stabilized.

cbartz (dismissed):
LGTM . Can you add the complexity label to the PR and provide a one-liner in the PR description? Thank you very much.

arturo-seijas (dismissed):

yanksyoon (dismissed):

weiiwang01 (approved):

gregory-schiano (approved):

PR #357, opened 2024-06-17
Add support for canonical k8s

Thanhphan1147:
This PR adds an input `use-canonical-k8s` to the `integration-test` workflow to allow deploying juju applications on canonical-k8s. 

For now, this parameter **must be used** in conjuction with `provider: k8s` so that `action-operator` bootstraps canonical-k8s as a generic cluster when creating the juju controller. 

`action-operator` will add support for canonical-k8s in the future. Until then, this will be used as a workaround since it's needed to test workflows running on canonical-k8s.

### Checklist


arturo-seijas (commented):
It looks like some tests are filing. Can you double check?

.github/workflows/integration_test_run.yaml
```diff
@@ -171,10 +175,26 @@ jobs:
         )) || inputs.runs-on
       }}
     steps:
+      - name: Setup Canonical k8s
+        if: ${{ inputs.use-canonical-k8s }} == true
+        id: canonical-k8s
+        run: |
+          sudo snap install k8s --channel=1.30-classic/beta --classic
```
arturo-seijas, line 182: I think you can reuse the `channel` input here currently used for microk8s to be able to customise the snap channel

cbartz:
@Thanhphan1147 what is the status here, please?

arturo-seijas (approved):

.github/workflows/integration_test_run.yaml
```diff
@@ -174,6 +178,21 @@ jobs:
         )) || inputs.runs-on
       }}
     steps:
+      - name: Setup Canonical k8s
+        if: inputs.use-canonical-k8s && inputs.provider == 'k8s'
+        id: canonical-k8s
+        run: |
+          sudo snap install k8s --channel=${{ inputs.channel }} --classic
```
javierdelapuente, line 185: Maybe use another input variable for the channel? The default is latest/stable, and that does not exists for that snap.

.github/workflows/integration_test_run.yaml
```diff
@@ -174,6 +178,21 @@ jobs:
...
+          sudo k8s bootstrap
+          mkdir -p ~/.kube
+          sudo k8s config > ~/.kube/config
+          echo "kubeconfig=$(sudo k8s config | base64 -w 0)" >> $GITHUB_OUTPUT
+          
+          IPADDR=$(ip -4 -j route get 2.2.2.2 | jq -r '.[] | .prefsrc')
+          LB_FIRST_ADDR="$(echo "${IPADDR}" | awk -F'.' '{print $1,$2,$3,100}' OFS='.')"
```
javierdelapuente, line 192: I do not know much about this, but you will this work out of the box? You are in a subnet with other machines that may reply to arp requests for some of those ips.

nrobinaubertin:
[CI](https://github.com/canonical/gateway-api-integrator-operator/pull/85) [is](https://github.com/canonical/gateway-api-integrator-operator/pull/86) [failing](https://github.com/canonical/gateway-api-integrator-operator/pull/84) for gateway-api. This is because the k8s cluster cannot be bootstrapped or because it is trying to install the stable channel which doesn't exist.
I write this comment as reference on those stalled PRs.

.github/workflows/integration_test.yaml
```diff
@@ -158,6 +158,10 @@ on:
           If this parameter is not specified, the default behavior is to use 'artifact' for pull requests 
           originating from forked repositories, and 'registry' for all other cases.
         default: ""
+      use-canonical-k8s:
```
yhaliaw, line 161: It might good to document this added feature in the docs/changelog.md

yanksyoon (approved):
would there be a test we could add to this change? LGTM!

yanksyoon (approved):

.github/workflows/integration_test.yaml
seb4stien: Why introducing a new variable as we already have "provider"?

.github/workflows/integration_test_run.yaml
Thanhphan1147: since this is a reusable workflow to run anywhere this is the simplest way I can think of to get a LB range, I'd assume cilium will at least try to see if the IP is available before taking it, so far from my testing it has been going fine, but I'll keep that in mind to see if this eventually becomes a problem.

.github/workflows/integration_test.yaml
Thanhphan1147: the problem with `provider` is that it can also be used to bootstrap a generic k8s cluster (if you install your own k8s prior to running juju bootstrap for example), not just `canonical-k8s`. I'd imagine action-operator will support a `canonical-k8s` provider in the future but currently we need both `provider: k8s` and `use-canonical-k8s: true`

seb4stien (approved):

.github/workflows/workflow_test.yaml
```diff
@@ -125,4 +125,20 @@ jobs:
...
+      extra-arguments: |
+        --kube-config=~/.kube/config
+      use-canonical-k8s: true
+      juju-channel: 3.5/stable
+      channel: edge
+      charmcraft-channel: latest/edge
+      # We need large runners to be able to deploy canonical k8s
+      self-hosted-runner: true
+      self-hosted-runner-image: noble
+      self-hosted-runner-label: large
```
arturo-seijas, line 144: Can't you use a small one here?

Thanhphan1147: canonical-k8s requires a large runner I think otherwise bootstrapping of the cluster will fail

arturo-seijas: It is not obvious for anyone reading the code. Please, next time add a comment explaining why a non obvious decision was made

PR #364, opened 2024-06-25
Skip upload resources if they have an upstream-source

weiiwang01:
Some resources are built outside of the `operator-workflows` and typically include an `upstream-source` field in their definitions. In the publishing workflows, these resources should be excluded from uploading, and the charm should be published directly.


cbartz (approved):
LGTM

Thanhphan1147 (approved):

PR #365, opened 2024-06-25
Add an option to select the series for self-hosted runners.

weiiwang01:
Add an option to select the series for self-hosted runners, as the current runner label construction does not allow selecting a runner that is not `jammy`


cbartz:
There is already a PR for this https://github.com/canonical/operator-workflows/pull/345

weiiwang01:
> @weiiwang01 There is already a PR for this #345

Sorry I missed that, I will close mine, thanks!

PR #371, opened 2024-07-03
Add a comment linking to the contributing guide on PR open

arturo-seijas:
Applicable spec: <link>

### Overview

Add a comment linking to the contributing guide on PR open

### Checklist

### DIFF
--- a/.github/workflows/comment_contributing.yaml
+++ b/.github/workflows/comment_contributing.yaml
@@ -1,0 +1,41 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+name: Comment on the pull request
+
+on:
+  workflow_call:
+
+jobs:
+  comment-on-pr:
+    name: Comment on PR
+    runs-on: ubuntu-22.04
+    if: github.event.workflow_run.event == 'pull_request'
+    steps:
+      - uses: actions/checkout@v4.1.7
+      - name: Comment on PR
+        uses: actions/github-script@v7.0.1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const fs = require('fs');
+            const comments = "Thanks for opening a pull request! Please, check that you are following our \
+            [contributing guide](https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md) \
+            and you are not introducing breaking changes in the target channel."
+            const owner = context.payload.workflow_run.repository.owner.login;
+            const repo = context.payload.workflow_run.repository.name;
+
+            const pull = (await github.rest.pulls.list({
+                owner: owner,
+                repo: repo,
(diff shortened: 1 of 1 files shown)


nrobinaubertin (dismissed):
Good idea !

arturo-seijas:
> Good idea !

It was @jdkandersson 's :)

.github/workflows/comment_contributing.yaml
```diff
@@ -0,0 +1,41 @@
...
+name: Comment on the pull request
+
+on:
+  workflow_call:
+
+jobs:
+  comment-on-pr:
+    name: Comment on PR
+    runs-on: ubuntu-22.04
+    if: github.event.workflow_run.event == 'pull_request'
```
weiiwang01, line 13: I think we can probably remove this condition and use the `pull_request_target` `open` event to trigger this workflow. This way, it will only create one comment per pull request. Additionally, we should probably only comment on pull requests from forks.

```
on:
  pull_request_target:
    types: [opened]

jobs:
  comment-contributing-guide:
    if: ${{ github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name }}
    uses: canonical/operator-workflows/.github/workflows/comment_contributing.yaml@main
    secrets: inherit
```

.github/workflows/comment_contributing.yaml
```diff
@@ -0,0 +1,41 @@
...
+            const owner = context.payload.workflow_run.repository.owner.login;
+            const repo = context.payload.workflow_run.repository.name;
+
+            const pull = (await github.rest.pulls.list({
+                owner: owner,
+                repo: repo,
+            })).data.filter(pr => pr.head.sha == context.payload.workflow_run.head_sha)[0];
+            const issue_number = pull.number;
+
+            const createComment = async (body) => {
```
yanksyoon, line 34: Should we call this?

yanksyoon (dismissed):
Minor question - not sure if declaration of arrow func gets called :)

.github/workflows/comment_contributing.yaml
weiiwang01: nvm, I think the comment above is outside the scope of this pull request, sorry.

weiiwang01: I think the `github.event.workflow_run.event == 'pull_request'` condition exists because the test workflow used to run in the `push` event as well (for publish charms). But that's probably not needed anymore.

arturo-seijas: Mmm makes sense. Updated

weiiwang01 (dismissed):

.github/workflows/comment_contributing.yaml
arturo-seijas: Oops

yanksyoon (approved):

weiiwang01 (approved):

Issue #441, opened 2024-10-08
Enable `security.nesting` for `rockcraft pack`


kelkawi-a:
Reference: https://github.com/canonical/craft-providers/issues/591

We have a rock to be built which spins up a docker container as part of the build process to generate some database queries and metadata. Such functionality is not enabled by default in `rockcraft`. After some digging around, it seems that setting `lxc --project=rockcraft profile set default security.nesting true` on the lxc container before running `rockcraft pack` allows this to be done. Would it be possible to add this setting as a parameter to the integration tests?

cbartz:
Seems this is now implemented by #443 .

PR #442, opened 2024-10-09
Add enable security nesting input to itest

kelkawi-a:
### Overview

Addresses #441 by adding an input parameter to the integration tests for enabling security nesting when building rocks.

### Rationale

This is needed for rocks with nested containers (e.g. docker containers) as part of the build process.

### Checklist


internal/build/action.yaml
```diff
@@ -26,6 +26,8 @@ inputs:
   rockcraft-repository:
     description: Pull and build rockcraft from source instead of using snapstore version (this means that the rockcraft-channel input will be ignored).
     default: ""
+  rockcraft-enable-security-nesting:
+    description: Set security.nesting=true on the rockcraft lxc project.
```
yanksyoon, line 30: Hello! Would you be able help briefly elaborate what security.nesting=True on lxc project does please?

src/build.ts
yanksyoon, line 1: I think you may want to build and push the build dists here as well, would you be able to run `npm run bundle` to do so please?

alithethird: I agree you should build and add it into the PR, I was not able to find a CI run that builds dists here.

internal/build/action.yaml
kelkawi-a: I've updated the description, [security nesting](https://discuss.linuxcontainers.org/t/what-does-security-nesting-true/7156/2) can be used to allow containers to spin up other containers. This is needed when building the [Airbyte rock](https://github.com/canonical/airbyte-k8s-operator) as it spins up a docker image inside the rock as part of the build process.

src/build.ts
kelkawi-a: Done :) though i'm not sure if I've handled the boolean casting correctly in the TS code, could you please let me know?

alithethird: This is a good concern. We need to make sure it's done correctly.

alithethird: I think you should use `getBooleanInput` according to https://github.com/actions/toolkit/tree/main/packages/core#inputsoutputs

alithethird (commented):
Please don't merge yet.

PR #470, opened 2024-11-20
Fix file resource builder github group text

addyess:
Fix a github label show while running the build on a file resource

image

this should read instead:

```
Build resource my-charm-resource
```


addyess:
@weiiwang01 i found a quick fix up on the previous pr #454

arturo-seijas (approved):

weiiwang01 (approved):

Issue #496, opened 2024-12-03
Craft tools get executed with an empty `$PATH`


lengau:
When [running craft tools](https://github.com/canonical/operator-workflows/blob/27022a81469d5c31da80f2e5fa26e7fa2bdb260f/src/build.ts#L107-L110), the environment gets replaced rather than updated, so snapd gets an [empty `$PATH`](https://bugs.launchpad.net/snapd/+bug/2090938).

This causes Charmcraft to think the cache directory is `/cache` because snapd sets `SNAP_USER_COMMON` to an empty value, which causes charmcraft to fail when trying to create the cache directory.

yanksyoon:
Operator workflows side has been fixed by: https://github.com/canonical/operator-workflows/pull/497.
Closing :)

PR #497, opened 2024-12-03
Extend the process env with new envvars rather than wholesale replacement

addyess:
Environmental variables from the runner should be extended rather than replaced in exec calls

BugFix for #496 

### Overview

actions/exec when called with `env` will replace all the environment variables  as evidenced [here](https://github.com/actions/toolkit/blob/main/packages/exec/src/toolrunner.ts#L357).  Instead, we should extend the existing processes env variables with our new settings.

### Rationale

Helps address missing `PATH` when calling `charmcraft` and other snap based tools.

### Workflow Changes

No high level changes

### Checklist

### DIFF
--- a/src/build.ts
+++ b/src/build.ts
@@ -106,7 +106,7 @@ async function buildCharm(params: BuildCharmParams): Promise<void> {
   core.startGroup('charmcraft pack')
   await exec.exec('charmcraft', ['pack', '--verbosity', 'trace'], {
     cwd: params.plan.source_directory,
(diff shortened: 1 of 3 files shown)


addyess:
@arturo-seijas and @weiiwang01 do you folks have thoughts on this one?

arturo-seijas (approved):
Thanks for the contribution :)

addyess:
The integration tests failed, but seem unreleated to my changes i think

weiiwang01 (approved):

PR #503, opened 2024-12-12
Update if conditions for Allure

swetha1654:
### Overview

Allure workflow fails when the workflow is re-run due to the mkdir command. Restricting the allure workflow to run only during the first attempt.

### Checklist

### DIFF
--- a/.github/workflows/allure_report.yaml
+++ b/.github/workflows/allure_report.yaml
@@ -10,7 +10,7 @@ jobs:
     name: Publish Allure report
     runs-on: ubuntu-latest
     timeout-minutes: 5
-    if: always() && !cancelled()
+    if: always() && !cancelled() && github.run_attempt == '1'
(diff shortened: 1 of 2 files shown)


yanksyoon (approved):
The change lgtm, but is it intended to ignore the runs after the first attempt?

swetha1654:
> The change lgtm, but is it intended to ignore the runs after the first attempt?

The allure reports are currently configured to store the test results of every run number. Storing the results for every run attempt under every run_number would require more tweaking of the code. I could either chose to store the last attempt overwriting everything or just keep the first attempt and i decided to do the latter.

jdkandersson (approved):

jdkandersson:
I think the first attempt makes sense because it shows us what went wrong the first time in case of re-runs. If we went with the last we might only see the successful ones after running retries

PR #527, opened 2025-01-14
Support testing with tox-uv for test speedups

addyess:
UV is a python package manager written in rust and it's FAR faster at installing packages that pip.  

By switching to tox-uv, package installs and tests like unit, static, lint, can run much faster.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -18,6 +18,11 @@ on:
         type: string
         description: The working directory for jobs
         default: "./"
+      python-version:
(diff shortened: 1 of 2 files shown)


addyess:
@arturo-seijas and @weiiwang01 

consider the unit-tests speeds ups between simple-uv and simple.  You'll noticed that uv can be up to 3x faster!

yanksyoon (approved):
Thank you for the contribution! LGTM!

arturo-seijas (approved):

PR #545, opened 2025-01-31
Ensure the check-libraries action runs on charm path

addyess:
### Overview
* Charms where the `/lib` directory is buried down in the repo can't reach this library check

### Details
* update the description of `charm-directory` because it's not  just about docs

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -8,7 +8,7 @@ on:
     inputs:
       charm-directory:
         type: string
-        description: The working directory for the docs
+        description: The working directory for the charm
         default: "."
       charmcraft-channel:
         description: Charmcraft channel to use
@@ -485,6 +485,7 @@ jobs:
         with:
           credentials: ${{ secrets.CHARMHUB_TOKEN }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
+          charm-path: ${{ inputs.working-directory }}/${{ inputs.charm-directory }}
           charmcraft-channel: ${{ inputs.charmcraft-channel }}
   required_status_checks:
     name: Required Test Status Checks
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -51,9 +51,13 @@ on:
         required: false
         default: false
 
+        
 concurrency:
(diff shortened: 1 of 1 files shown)


addyess:
I'm fixing many of these errors over on https://github.com/charmed-kubernetes/actions-operator/pull/93

.github/workflows/test.yaml
```diff
@@ -410,9 +414,8 @@ jobs:
         uses: actions/upload-artifact@v4
         if: always() && github.event_name == 'pull_request' && !cancelled()
         with:
-          name: report
+          name: report-${{ env.CONCURRENCY_GROUP}}
           path: report.json
-          overwrite: true
```
addyess, line 415: Not sure if this is something you'd actually like -- but with calling this thing in a matrix style, each report kinda needs its own name.  The "overwrite:true" is a little flaky on github's end

.github/workflows/test.yaml
```diff
@@ -51,10 +51,15 @@ on:
         required: false
         default: false
 
+        
 concurrency:
   group: operator-workflows-${{ github.workflow }}-${{ inputs.working-directory }}-tests-${{ github.ref }}-self-hosted-${{ inputs.self-hosted-runner }}-py${{ inputs.python-version }}-uv-${{ inputs.with-uv }}
```
arturo-seijas, line 56: Can't we use the env var `CONCURRENCY_GROUP` here to avoid duplication? Not sure if is supported but I think it's worth a try

addyess: I struggled with this too... but it doesn't seem you can use env vars in a concurrency group.  that's why i put them so close together in the file

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#concurrency

> Use concurrency to ensure that only a single job or workflow using the same concurrency group will run at a time. A concurrency group can be any string or expression. The expression can only use [github](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context), [inputs](https://docs.github.com/en/actions/learn-github-actions/contexts#inputs-context) and [vars](https://docs.github.com/en/actions/learn-github-actions/contexts#vars-context) contexts. For more information about expressions, see [Evaluate expressions in workflows and actions](https://docs.github.com/en/actions/learn-github-actions/expressions).

arturo-seijas: Somehow, I'm not surprised. Thanks for the PR :)

arturo-seijas (approved):

javierdelapuente (approved):

Issue #549, opened 2025-02-05
comment on prs fails to find coverage repo


addyess:
#544 introduced a regression where the workflows cannot comment on  PRs with the unit-test reports. 

here's an [example](https://github.com/canonical/k8s-operator/actions/runs/13159233660/job/36723521827) of a failed run

PR #559, opened 2025-02-12
fix: split artifact filtering and commenting

addyess:
Applicable spec: Issue: #549 

### Overview

As part of fixing the PR commenter, split apart the filtering from the downloading of artifacts

### Rationale

At the moment, we download artifacts that aren't available to the comment job.
We should instead let each "comment" step download and comment together. 

### Workflow Changes

Each report will now generate a separate set of comments on a PR. The sha will appear in the header to help remove old comments if new commits are made for a PR. 

### Checklist

- [ ] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied

### DIFF
--- a/.github/workflows/comment.yaml
+++ b/.github/workflows/comment.yaml
@@ -32,8 +32,8 @@ jobs:
                 repo: context.payload.workflow_run.repository.name,
                 run_id: ${{ github.event.workflow_run.id }},
             });
-            const zips = artifacts.data.artifacts.filter(artifact =>
-              artifact.name.startsWith('${{ env.ARTIFACT_NAME }}-')
+            const zips = artifacts.data.artifacts.flatMap(artifact =>
+              artifact.name.startsWith('${{ env.ARTIFACT_NAME }}-')? [artifact.name] : []
             );
             return JSON.stringify(zips);
   comment-on-prs:
--- a/.github/workflows/comment.yaml
+++ b/.github/workflows/comment.yaml
@@ -13,57 +13,44 @@ env:
   ARTIFACT_NAME: "report"
 
 jobs:
-  download-artifacts:
-    name: Download Artifacts
+  filter-artifacts:
+    name: Filter Artifacts
     runs-on: ubuntu-24.04
     if: github.event.workflow_run.event == 'pull_request'
     outputs:
-      artifacts: ${{ fromJson(steps.download-artifact.outputs.result) }}
+      artifacts: ${{ fromJson(steps.filter-artifacts.outputs.result) }}
     steps:
       - uses: actions/checkout@v4.2.2
-      - id:   download-artifact
-        name: Download test report artifact
+      - id:   filter-artifacts
+        name: Filter test report artifacts
         uses: actions/github-script@v7.0.1
         with:
           script: |
             const fs = require('fs');
-            const owner = context.payload.workflow_run.repository.owner.login;
-            const repo = context.payload.workflow_run.repository.name;
-            const run_id = ${{ github.event.workflow_run.id }};
             const artifacts = await github.rest.actions.listWorkflowRunArtifacts({
-                owner: owner,
-                repo: repo,
-                run_id: run_id,
+                owner: context.payload.workflow_run.repository.owner.login,
+                repo: context.payload.workflow_run.repository.name,
+                run_id: ${{ github.event.workflow_run.id }},
             });
-            const report_artifacts = artifacts.data.artifacts.filter(artifact =>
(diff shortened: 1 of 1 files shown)


.github/workflows/comment.yaml
```diff
@@ -13,57 +13,44 @@ env:
...
-                    owner: owner,
-                    repo: repo,
-                    artifact_id: artifact.id,
-                    archive_format: 'zip',
-                });
-                const zipPath = `${artifact}.zip`;
-                fs.writeFileSync(zipPath, Buffer.from(download.data));
-                return zipPath;
-              })
             );
```
addyess, line 37: We can stop downloading in this job, a new runner will spin up to create the PR comments

.github/workflows/comment.yaml
```diff
@@ -13,57 +13,44 @@ env:
...
     steps:
-      - name: Unzip
-        shell: bash
-        run: unzip ${{ matrix.artifact }}
+      - name: Download artifact
+        uses: actions/download-artifact@v4
+        with:
+          repository: ${{ github.event.workflow_run.repository.name }}
+          run-id: ${{ github.event.workflow_run.id }}
+          name: ${{ matrix.artifact }}
```
addyess, line 53: Downloading the artifact is made simpler by using the action.  We know the `report.json` file will be automatically put into place on the runner

.github/workflows/comment.yaml
```diff
@@ -96,7 +85,9 @@ jobs:
                     issue_number: issue_number,
                 });
                 const githubActionsComments = existingComments.data.filter(
-                    comment => comment.user.login == 'github-actions[bot]'
+                    comment =>
+                      comment.user.login == 'github-actions[bot]' &&
+                      !comment.body.startsWith(header),
```
addyess, line 90: here is that filtering.   We want to delete comments if they do not apply to this commit.

addyess:
@arturo-seijas or @weiiwang01 kindly would you mind a review?

.github/workflows/comment.yaml
```diff
@@ -13,57 +13,44 @@ env:
...
     strategy:
       matrix:
-        artifact: ${{fromJson(needs.download-artifacts.outputs.artifacts)}}
+        artifact: ${{fromJson(needs.filter-artifacts.outputs.artifacts)}}
     steps:
-      - name: Unzip
-        shell: bash
-        run: unzip ${{ matrix.artifact }}
+      - name: Download artifact
+        uses: actions/download-artifact@v4
```
arturo-seijas, line 49: Can you pin this to a specific version so that builds are closer to reproducible?

addyess: Done

arturo-seijas (approved):
Thanks!

.github/workflows/comment.yaml
```diff
@@ -13,57 +13,44 @@ env:
...
-                repo: repo,
-                run_id: run_id,
+                owner: context.payload.workflow_run.repository.owner.login,
+                repo: context.payload.workflow_run.repository.name,
+                run_id: ${{ github.event.workflow_run.id }},
             });
-            const report_artifacts = artifacts.data.artifacts.filter(artifact =>
+            const zips = artifacts.data.artifacts.filter(artifact =>
               artifact.name.startsWith('${{ env.ARTIFACT_NAME }}-')
-            )
```
weiiwang01, line 40: I think this will make the output a list of artifact objects instead of a list of artifact names, which I suppose is the intended input for the `comment-on-prs` job?

.github/workflows/comment.yaml
weiiwang01: I think the problem here is that for multiple results, as they are all running on the same commit and thus share the same commit SHA. As a result, multiple `comment-on-prs` jobs will overwrite each other, leaving only one comment in the pull request thread.

I would try adding the name of the result to the header so they won't overlap with each other.

weiiwang01 (approved):
@addyess sorry for the delay, overall looks good to me, thanks!

.github/workflows/comment.yaml
addyess: I agree @weiiwang01 .  If there are multiple reports on a PR, we want to delete comments if they originated from a previous commit id. I know that since this is running in a matrix, we want to skip deleting a comment if it originates from the same sha256, hence the `!comment.body.startsWith(header)`

I envision that if there are 3 reports, there will be 3 * N comments showing up in the report -- all with the same header associated with that commit id (N being the list of comments in the report).

weiiwang01: Yes, I think you are right.

PR #576, opened 2025-02-24
Correct workflow-runner organization which is missing

addyess:
Applicable spec: #549 

### Overview

workflow runs offer an unset organization
example) https://github.com/canonical/k8s-operator/actions/runs/13507118728/job/37739162857

### Rationale

Appropriately set the organization when detected pull requests

### Workflow Changes

None

### Checklist

### DIFF
--- a/.github/workflows/comment.yaml
+++ b/.github/workflows/comment.yaml
@@ -38,6 +38,7 @@ jobs:
         parts=(${REPO_FULL_NAME//\// })
         echo "ORG=${parts[0]}" >> $GITHUB_OUTPUT
         echo "REPO=${parts[1]}" >> $GITHUB_OUTPUT
+        echo "REPO_FULL_NAME=${REPO_FULL_NAME}" >> $GITHUB_OUTPUT
         echo "SHA=${{ github.event.pull_request.head.sha }}" >> $GITHUB_OUTPUT
         echo "IS_PR=${{ github.event_name == 'pull_request' }}" >> $GITHUB_OUTPUT
     - name: Workflow run setup
--- a/.github/workflows/comment.yaml
+++ b/.github/workflows/comment.yaml
@@ -45,7 +45,7 @@ jobs:
       if: github.event_name == 'workflow_run'
       run: |
         echo "RUN_ID=${{ github.event.workflow_run.id }}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


addyess:
the example workflow run indicates something is not right with the `download-artifacts` when it filters by pattern `report-*` as it finds 0 artifacts to download.

.github/workflows/comment.yaml
```diff
@@ -60,9 +61,10 @@ jobs:
       - name: Download artifacts
         uses: actions/download-artifact@v4.1.8
         with:
+          pattern: "${{ env.ARTIFACT_NAME }}-*"
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
addyess, line 65: The GH token is required for `workflow-runs` according to [download-artifacts](https://github.com/actions/download-artifact/tree/v4.1.8?tab=readme-ov-file#inputs) inputs

addyess:
@weiiwang01 so close i can taste it now.  Thanks for permissions to test the workflows.  that's been a huge help

.github/workflows/comment.yaml
```diff
@@ -38,14 +38,15 @@ jobs:
...
+        echo "REPO_FULL_NAME=${REPO_FULL_NAME}" >> $GITHUB_OUTPUT
         echo "SHA=${{ github.event.pull_request.head.sha }}" >> $GITHUB_OUTPUT
         echo "IS_PR=${{ github.event_name == 'pull_request' }}" >> $GITHUB_OUTPUT
     - name: Workflow run setup
       id: run
       if: github.event_name == 'workflow_run'
       run: |
         echo "RUN_ID=${{ github.event.workflow_run.id }}" >> $GITHUB_OUTPUT
-        echo "ORG=${{ github.event.workflow_run.organization.login }}" >> $GITHUB_OUTPUT
+        echo "ORG=${{ github.event.workflow_run.repository.owner.login }}" >> $GITHUB_OUTPUT
```
yhaliaw, line 49: Does the workflow still work correctly if the repository owner is not an organization?
Would it be better to rename the `ORG` to `OWNER`?

yhaliaw (commented):
One question on the naming of the variable.

.github/workflows/comment.yaml
addyess: good catch, lets' go!  thanks @yhaliaw

weiiwang01 (approved):

yanksyoon (approved):

PR #627, opened 2025-04-29
Update `promote charm` logic

swetha1654:
Applicable spec: <link>

### Overview

- Update `promote_charm` workflow logic to use `charmcraft status` to obtain base information instead of `charmcraft.yaml`. The new updated logic finds the first available base information under the track to be promoted and sets that as the default base information if user-provided base inputs are not present.

### Rationale

Current promote_charm workflow assumes that charmcraft.yaml has a bases section. This causes issues when the workflow is called with custom inputs for `base_name`, `base_channel` and `base-architecture`.

### Checklist


weiiwang01:
Thanks for this update, @swetha1654! But when I was reviewing your code, I thought maybe the entire workflow needs to be reworked instead of just patching this single problem.

Currently, the workflow extracts the available charm bases (a combination of Ubuntu release and architecture) from the `charmcraft.yaml` and then promotes the charm based on that. However, as the workflow is intended to promote the charm that's already released on charmhub, I think we should only use the name information from `charmcraft.yaml` and obtain the available bases for the charm from charmhub (using the `charmcraft status <charm-name> --format json` command), and promote revisions of all bases to the new channel. This should be more aligned with the fact that more and more of our charms are becoming multi-based.

I'm not sure if my description is clear, if not, I can create a pull request for this purpose instead.

swetha1654:
@weiiwang01 makes sense to me, let me update the workflow accordingly. Thanks!

.github/workflows/promote_charm.yaml
```diff
@@ -78,96 +60,102 @@ jobs:
...
+          if [[ ! -f "charmcraft.yaml" ]]; then
+            echo "Error: charmcraft.yaml not found in the current directory."
+            exit 1
+          fi
 
-          CHARMCRAFT_YAML_PATH=${{ inputs.working-directory }}/${{ inputs.charm-directory }}/charmcraft.yaml
-          if [ ! -f $CHARMCRAFT_YAML_PATH ]; then
-              echo "charmcraft.yaml not found in $CHARMCRAFT_YAML_PATH"
-              exit 1
+          CHARM_NAME=$(yq e '.name' charmcraft.yaml)
```
arturo-seijas, line 87: Why are you removing the charm-directory arg here? Note that the charmcraft.yaml files are not always in the root of the repository

arturo-seijas (commented):
Can you open a PR for each individual change?

.github/workflows/promote_charm.yaml
swetha1654: I haven't removed it entirely, I added the `working-directory` field instead of setting the `CHARMCRAFT_YAML_PATH` as i thought that would be a neater way to do it.

seb4stien (approved):

.github/workflows/promote_charm.yaml
arturo-seijas: Oh, I understand now. Thanks

swetha1654:
> Can you open a PR for each individual change?

May I know the reason to open a PR for each individual change @arturo-seijas ? I wanted to keep them in one PR so that all the promote charm changes remain in one place and it will be easier to revert if any backward compatibility issues arise.

arturo-seijas:
> > Can you open a PR for each individual change?
> 
> May I know the reason to open a PR for each individual change @arturo-seijas ? I wanted to keep them in one PR so that all the promote charm changes remain in one place and it will be easier to revert if any backward compatibility issues arise.

A smaller PR has several advantages:
- Smaller changes are less risky than bigger ones
- Easier to review
- Easier to test and to revert if things go south
- Will also result in shorter delivery lead time
- 
And in this case, I see the changes are, although related, independent, so it's convenient to split it

swetha1654:
> > > Can you open a PR for each individual change?
> > 
> > 
> > May I know the reason to open a PR for each individual change @arturo-seijas ? I wanted to keep them in one PR so that all the promote charm changes remain in one place and it will be easier to revert if any backward compatibility issues arise.
> 
> A smaller PR has several advantages:
> 
> * Smaller changes are less risky than bigger ones
> * Easier to review
> * Easier to test and to revert if things go south
> * Will also result in shorter delivery lead time
> 
> And in this case, I see the changes are, although related, independent, so it's convenient to split it

@arturo-seijas , makes sense. Let me split it.

arturo-seijas (approved):

PR #629, opened 2025-05-06
feat: add label pr approval workflow

cbartz:
Applicable spec: <link>

### Overview

Add a workflow that approves a PR based on existence of a certain label.

Tested [here](https://github.com/canonical/github-runner-webhook-router/pull/71).

### Rationale

Github does not support adapting approval rules based on labels. In order to be less strict for certain PR's, it make sense to add one approval (e.g. based on "trivial" label).

### Workflow Changes

n/a

### Checklist


.github/workflows/label_pr_approval.yaml
```diff
@@ -0,0 +1,53 @@
...
+            const pull = (await github.rest.pulls.get({
+              owner: owner,
+              repo: repo,
+              pull_number: context.payload.number,
+            })).data;
+
+            let found = false;
+            for (const prLabel of pull.labels) {
+              if (labels.includes(prLabel.name)) {
+                console.log("Found a matching label '" + prLabel.name + "', will approve PR");
```
weiiwang01, line 39:
```suggestion
                console.log(`Found a matching label '${prLabel.name}', will approve PR`);
```

.github/workflows/label_pr_approval.yaml
```diff
@@ -0,0 +1,53 @@
...
+      labels:
+        description: List of labels in JSON format, i.e. '["trivial", "documentation"]', to indicate that a PR should be approved. When a pull request contains at least one of those labels, the PR will be approved.
+        type: string
+        default: '["trivial"]'
+
+jobs:
+  approve_pr:
+    name: Approve PR for matching label
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
```
weiiwang01, line 18:
```suggestion
    if: github.event_name == 'pull_request' && github.event.pull_request.base.repo.full_name == github.event.pull_request.head.repo.full_name
```

I think we should probably disable this workflow for pull requests initiated from forks, as running it from a fork may fail.

weiiwang01 (approved):

arturo-seijas (changes requested):
This effectively means that the PR opener can, by adding the appropriate tag, require only one approval, rendering the requirement for 2 approvals useless.

Instead of this, I suggest changing the repository configuration to require one approval instead of two for any scenario, since this change means that we won't be able to enforce 2 approvals before merging

javierdelapuente (approved):

cbartz:
@arturo-seijas 

> This effectively means that the PR opener can, by adding the appropriate tag, require only one approval, rendering the requirement for 2 approvals useless.
> 
> Instead of this, I suggest changing the repository configuration to require one approval instead of two for any scenario, since this change means that we won't be able to enforce 2 approvals before merging

If there are doubts, I suggest removing the usage of the workflow from this particular repository by removing `trivial_pr_approval`.

This functionality still makes sense for other repositories where most engineers are on the same team. Most PRs would not be trivial and would still require two approvals. If engineers are not acting responsibly with the label, then we have a deeper issue anyway.

yhaliaw (approved):

cbartz:
Closing this PR as there is not enough agreement inside the PFE team for this feature.

PR #684, opened 2025-06-16
Call internal workflows locally rather than with `@main`

### LINKED ISSUE #683: Fixed commits fail to publish charms
Notice how our release-branch CI needs to point to a fixed commit of the operator-workflows
Here:
https://github.com/canonical/k8s-operator/blob/release-1.32/.github/workflows/publish-charms.yaml#L44
```yaml
  publish-to-edge:
    needs: [configure-channel, charmcraft-channel]
    uses: canonical/operator-workflows/.github/workflows/publish_charm.yaml@e848763f1e14d3169781fe95b4b896399a1a8a92
    strategy:
      matrix:
        charm:
        - { path: ./charms/worker/k8s/, tagPrefix: k8s }
        - { path: ./charms/worker/,     tagPrefix: k8s-worker }
        arch:
        - amd64
        - arm64
```

the publish_charm at that commit
https://github.com/canonical/operator-workflows/blob/e848763f1e14d3169781fe95b4b896399a1a8a92/.github/workflows/publish_charm.yaml#L77

references `@main`
```yaml
  publish-charm:
    name: Publish charm to ${{ inputs.channel || needs.select-channel.outputs.destination-channel }}
    runs-on: ubuntu-latest
    needs: [select-channel]
    outputs:
      charm-directory: ${{ steps.publish.outputs.charm-directory }}
    steps:
      - uses: actions/checkout@v4.2.2
      - run: sudo snap install charmcraft --channel ${{ inputs.charmcraft-channel }} --classic
        # required to use skopeo embedded within rockcraft - otherwise a docker error
        # "io: read/write on closed pipe" will happen
      - run: sudo snap install rockcraft --classic
      - uses: canonical/operator-workflows/internal/publish@main
```

rather than 
```yaml
      - uses: canonical/operator-workflows/internal/publish@e848763f1e14d3169781fe95b4b896399a1a8a92
```

This basically means you cannot use PINNED versions of the `operator-workflow` b/c its internally referencing workflows that are unpinned from a commit.

addyess:
Fixes #683 

Avoids calling `canonical/operator-workflows/internal/<action>@main` because it prevents the operatork-workflows from being pinned.

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -153,10 +153,10 @@ on:
       upload-image:
         type: string
         description: >-
-          Can be either 'artifact' or 'registry', which determines the method by which images used 
-          in integration tests are uploaded. If set to 'artifact', the images will be uploaded as GitHub 
-          action artifacts. If set to 'registry', the images will be uploaded to ghcr. 
-          If this parameter is not specified, the default behavior is to use 'artifact' for pull requests 
+          Can be either 'artifact' or 'registry', which determines the method by which images used
+          in integration tests are uploaded. If set to 'artifact', the images will be uploaded as GitHub
+          action artifacts. If set to 'registry', the images will be uploaded to ghcr.
+          If this parameter is not specified, the default behavior is to use 'artifact' for pull requests
           originating from forked repositories, and 'registry' for all other cases.
         default: ""
       use-canonical-k8s:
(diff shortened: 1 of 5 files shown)


addyess:
@weiiwang01 @arturo-seijas  WDYT?

weiiwang01:
> @weiiwang01 @arturo-seijas WDYT?

I tried this before, and it doesn't work due to the limitations of reusable workflows. For example, if I use this workflow in another repository, such as `canonical/test`, the reference in the workflow, for instance, `uses: ./internal/get-plan`, will point to `canonical/test/internal/get-plan`, which doesn't exist, and the workflow will fail.

addyess:
okay @arturo-seijas and @weiiwang01 

I've checked out the workflows into `_workflow_repo` then called them with `./_workflow_repo/internal/<whatever>`

This works for this PR, so as promised i'll try from a REAL caller

PR #686, opened 2025-06-18
Fix lib publishing job

arturo-seijas:
Applicable spec: <link>

### Overview

Fixes a bug getting the path for the libs introduced in #679 

### Checklist

- [x] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [x] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)
- [x] The [changelog](`../docs/changelog.md`) is updated with user-relevant changes in the format of [keep a changelog v1.1.0](https://keepachangelog.com/en/1.1.0/)

PR #689, opened 2025-06-18
Optionally publish docs and libs

addyess:
### Overview
* alphabetize the publish_charm inputs
* add input `publish-docs` and `publish-libs`

### Reasons
Because of broken lib publishing discussed [here](https://github.com/canonical/operator-workflows/pull/686#issuecomment-2984373394), let's just optionally disable those so i can publish charms that have no lib changes.  I can publish lib changes manually if i need to

### DIFF
--- a/.github/workflows/publish_charm.yaml
+++ b/.github/workflows/publish_charm.yaml
@@ -10,27 +10,6 @@ on:
         type: string
         required: false
         description: 'Destination Channel'
-      identifier:
-        type: string
-        description: >-
-          The matching integration test workflow job identifier, see identifier in the integration test workflow input.
-        default: ""
-      resource-mapping:
-        type: string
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_charm.yaml
```diff
@@ -10,46 +10,58 @@ on:
...
+        description: >-
+          Whether to draft publish the documentation to Discourse.
+          If set to false, the documentation will not be published.
+        default: true
+      publish-libs:
+        type: boolean
+        description: >-
+          Whether to release charm libraries.
+          If set to false, the libraries will not be released.
+        default: true
```
addyess, line 41: :new: these two inputs are new

arturo-seijas (approved):

PR #740, opened 2025-08-01
Fix vale workflow

swetha1654:
Applicable spec: <link>

### Overview

Currently, the vale workflow is pointing to an older pinned version of the pracepta workflow causing a difference in the local vale runs and the workflow output. Moreover, the workflow did not have `nofilter` enabled, disabling this resulted in the vale checks running only on the new changes introduced in the PR rather than all the documentation in the repository. 

This is the reason vale workflows in the CI are always successful regardless of errors in the documentation.

### Checklist


yanksyoon (commented):
Thanks for the PR!
Would you have any test runs for the charm repositories that:
1. Have the documentation files/dirs that is set as default (/docs, readme, ...)
2. Do not have the default documentation files/dirs

Or perhaps even better, would there be a way for us to add these cases as a test in this repository?

weiiwang01 (approved):

swetha1654:
@yanksyoon I have updated the PR to test the workflow with the `test-upload-charm` in this repository itself. That would be an example of documentation that does not use the default files.
To provide an example that uses the default files, i have [tested](https://github.com/canonical/opencti-operator/actions/runs/17060681511/job/48366770526?pr=41) it on OpenCTI.

.github/workflows/test.yaml
```diff
@@ -134,12 +138,25 @@ jobs:
       - run: |
           sudo apt-get update -y && sudo apt-get install -y tree ruby-rubygems docutils
       - name: Install styles
-        uses: canonical/praecepta@1.0
+        uses: canonical/documentation-style-guide@main
+      - name: Resolve Vale file paths
+        id: resolve_vale_files
+        uses: actions/github-script@v7
```
arturo-seijas, line 144: Can you pin this action to 7.0.1?

srbouffard (approved):

arturo-seijas (approved):

addyess:
haha lovely!  All my PRs are broken now 🗡️

swetha1654:
@addyess ideally your PRs should not have failed since the vale action fails only if `vale_style_check` is enabled. I took a look at your repository and i see its not enabled but Vale still fails anyway. I did a bit of digging and found that this is an existing issue in the upstream action - https://github.com/errata-ai/vale-action/issues/89. I've raised a PR to skip the workflow entirely if vale_style_check is disabled - #767

PR #790, opened 2025-09-02
Add break-system-packages only if self-hosted-runner is true

amandahla:
Applicable spec: <link>

### Overview

Fix workflow for adding parameter only if self-hosted-runner is true and image is noble.

Example:
https://github.com/canonical/synapse-operator/actions/runs/17407525292/job/49415825462?pr=870

### Checklist


yanksyoon:
I think this would be the case from noble images and onwards :D

alithethird (approved):

PR #870, opened 2025-10-30
feat: mattermost ping

yanksyoon:
Applicable spec: <link>

### Overview

- Allows workflow triggers to get Mattermost ping after the integration test run.

### Rationale

- So that the user does not need to keep checking on whether the integration test has finished (e.g. waiting for tmate)

### Checklist

Working screenshot proof: 

image

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 Each revision is versioned by the date of the revision.
 
+## 2025-10-31
+- Add `mattermost-message` option that enables pinging a Matter most channel with a custom message.
+
 ## 2025-10-28
 - Fix breaking integration tests for VM charms when INTEGRATION_TEST_SECRET_ENV_NAME variables are not set.
 


.github/workflows/integration_test.yaml
```diff
@@ -53,6 +53,15 @@ on:
         type: string
         description: Command line arguments for the load test execution
         default: ""
+      mattermost-message:
+        type: string
+        required: false
+        description: |
+          Requires MM_WEBHOOK_URL secret value to be configured on the repository, and
```
swetha1654: is this up to the individual repo owners to configure this URL on their respective owners or will there be a follow up PR in [platform-engineering-repos-secrets](https://github.com/canonical/platform-engineering-repos-secrets) to configure this?

yanksyoon: Individual repo owners should update it as required, because we dont have a shared channel to do this at the moment!

.github/workflows/integration_test_run.yaml
```diff
@@ -427,6 +436,16 @@ jobs:
         with:
           name: allure-results-${{ env.UNIQUE_ARTIFACT_SUFFIX }}
           path: ${{ inputs.working-directory }}allure-results/
+      - name: Ping Mattermost
+        if: ${{ always() && inputs.mattermost-message }}
+        uses: mattermost/action-mattermost-notify@master
```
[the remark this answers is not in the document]
yanksyoon: Yeah Im not quite sure such process exists - but this workflow is from official mattermost repository which seems trustworthy

florentianayuwono (approved):
gonna love this feature, thank you!!

.github/workflows/integration_test_run.yaml
```diff
@@ -427,6 +436,16 @@ jobs:
...
+        with:
+          MATTERMOST_WEBHOOK_URL: ${{ secrets.MM_WEBHOOK_URL }}
```
Aldebarran: I suggest to use "MATTERMOST_WEBHOOK_URL" instead "MM_WEBHOOK_URL" for the secret name. Unless there's a technical reason to shorten it, I would avoid acronyms.

.github/workflows/integration_test_run.yaml
```diff
@@ -427,6 +436,16 @@ jobs:
         with:
           name: allure-results-${{ env.UNIQUE_ARTIFACT_SUFFIX }}
           path: ${{ inputs.working-directory }}allure-results/
+      - name: Ping Mattermost
```
Aldebarran: I suggest to go for "Notify Mattermost" which may be more explicit.

.github/workflows/integration_test.yaml
Aldebarran:
```suggestion
          Requires MATTERMOST_WEBHOOK_URL secret value to be configured on the repository, and
```

.github/workflows/integration_test.yaml
```diff
@@ -53,6 +53,15 @@ on:
...
+          `secrets: inherit` is configured on the workflow calling the reusable workflow.
+          Mattermost message contents to send. To ping relevant users, use Mattermost handles as
```
seb4stien:
```suggestion
          Mattermost message contents to send. To notify relevant users, use Mattermost handles as
```

.github/workflows/integration_test_run.yaml
seb4stien:
```suggestion
      - name: Notify Mattermost
```

.github/workflows/integration_test_run.yaml
```diff
@@ -427,6 +436,16 @@ jobs:
...
+        if: ${{ always() && inputs.mattermost-message }}
+        uses: mattermost/action-mattermost-notify@master
+        with:
+          MATTERMOST_WEBHOOK_URL: ${{ secrets.MM_WEBHOOK_URL }}
+          PAYLOAD: |-
+            {
+              "text": "[${{ github.repository }}](https://github.com/${{github.repository}}/actions/runs/${{ github.run_id }}) ${{inputs.mattermost-message}}",
+              "icon": "https://cdn3.iconfinder.com/data/icons/system-basic-vol-4-1/20/icon-note-attention-alt3-512.png"
```
seb4stien: Can we bring the icon to a place we control?

.github/workflows/integration_test_run.yaml
arturo-seijas: Can you pin this to a tag?

seb4stien:
Discussing with @javierdelapuente, we were wondering if the use case would involve editing the CI configuration before and after a run?
If so, we were wondering if an alternative could/would be to run a script locally to monitor the job?
Something like:
1. Fetching the workflow ID: `gh run -R canonical/ubuntu-motd-server-operator list -s in_progress --json databaseId`
2. Polling the status via something like: `gh run -R canonical/ubuntu-motd-server-operator view 18935433457`

However I'm not sure we would be able to capture the start of the tmate session.

seb4stien:
The main benefit of local polling would be to avoid to have to modify the CI of the project to receive notifications. As we can enable tmate by clicking the "debug" option when re-running a job, we could just start the script locally to wait for the tmate session to be ready.
I tried a quick POC that seems to do the job:
```bash
#!/bin/bash

set -euo pipefail

ACTIVE_RUN_ID=$(gh run list --status in_progress --json databaseId -q .[].databaseId)
[ -z "$ACTIVE_RUN_ID" ] && echo "Could not find any workflow running" && exit 1

while true; do
        JOBS_STATUS=$(gh run view $ACTIVE_RUN_ID --json jobs -q '.jobs[].steps[] | select(.name == "Tmate debugging session (self-hosted)") | .status')
        echo "[$(date +%H:%M:%S)] Job status: $JOBS_STATUS"

        if echo $JOBS_STATUS | grep -q completed; then
                echo "Tmate session is ready"
                notify-send "Tmate ready"
                break
        fi

        sleep 10
done
```

PR #884, opened 2025-11-14
Parallelize the build process

weiiwang01:
Applicable spec: <link>

### Overview

Run the build process in parallel with the setup process during integration test runs.

During the setup phase (installing Juju, microk8s, and bootstrapping Juju), build artifacts (charm rock, etc.) are not required. Therefore, the build can proceed in parallel with the setup phase.

### Checklist

### DIFF
--- a/.github/workflows/integration_test_run.yaml
+++ b/.github/workflows/integration_test_run.yaml
@@ -372,7 +372,7 @@ jobs:
 
       - run: sudo apt install skopeo -y
       - name: Plan Integration
-        uses: canonical/operator-workflows/internal/plan-integration@parallel-build
+        uses: canonical/operator-workflows/internal/plan-integration@main
         id: plan-integration
         with:
           plan: ${{ inputs.plan }}


weiiwang01:
@copilot Come up with a changelog for this pull request and reply to me here:

src/plan-integration.ts
```diff
@@ -9,16 +9,92 @@ import path from 'path'
...
+        core.info(`[${job.status.toUpperCase()}] ${job.name}`)
+      }
+    }
+    if (targetJobs.length === successes) {
+      return
+    }
+  }
+}
+
+async function downloadArtifact(
```
arturo-seijas: I thought this logic was already implemented somewhere else in the pipeline

arturo-seijas (commented):
What are all the changes in the "dist" directory? Can you push those separately if needed?

weiiwang01:
> What are all the changes in the "dist" directory? Can you push those separately if needed?

They are the compiled results of the source code and are needed for the action to run. The file changes are most likely because renovate bumped the dependency but did not run the compilation process, so the files haven't been updated for a while.

src/plan-integration.ts
```diff
@@ -9,16 +9,92 @@ import path from 'path'
...
+import * as github from '@actions/github'
+
+function sleep(ms: number) {
+  return new Promise(resolve => setTimeout(resolve, ms))
+}
+
+async function waitBuild(githubToken: string): Promise<void> {
+  const octokit = github.getOctokit(githubToken)
+
+  while (true) {
```
arturo-seijas: Maybe we should cap the time here

src/plan-integration.ts
weiiwang01: Unfortunately, no. Usually when an artifact download fails, the workflow simply errors out because downloads rarely fail. But in this case it needs to be retried due to a race condition caused by not using the official job-dependencies system.

src/plan-integration.ts
```diff
@@ -9,16 +9,92 @@ import path from 'path'
...
+        run_id: github.context.runId,
+        attempt_number: github.context.runAttempt,
+        per_page: 100
+      }
+    )
+    const targetJobs = jobs.filter(j => (j.name || '').includes('Build'))
+    if (targetJobs.length === 0) {
+      core.info('no build jobs')
+      return
+    }
```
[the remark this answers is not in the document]
weiiwang01: I think that's highly unlikely, because when the waiting occurs after setup the run has already been running for five minutes. Also, when a job is queued it is still accessible via the API, so it will also be subject to the wait.

PR #985, opened 2026-04-03
Depreciated Jira workflow. Use https://github.com/canonical/gh-jira-sync-bot instead

seb4stien:
Platform Engineering has fully transitioned to https://github.com/canonical/gh-jira-sync-bot which is reacher in terms of features.

Based on this search, there are two external repositories that are referencing it because they are based on PFE charm template but they are not very active and most likely never configured it https://github.com/search?q=org%3Acanonical+canonical%2Foperator-workflows%2F.github%2Fworkflows%2Fjira.yaml&type=code


arturo-seijas (commented):
This effectively removes the workflow, which seems to contradict with the message

arturo-seijas (approved):

PR #997, opened 2026-04-13
Fix/terraform tests with lxd and k8s

seb4stien:
### Overview

Use a single-controller when both lxd and k8s providers are requested.

### Rationale

To simplify tests, it's easier to have a single controller.

### Workflow Changes

- Boostrap lxd first if requested.
- Add k8s as a cloud to lxd if requested.

Tested on https://github.com/canonical/wazuh-server-operator/actions/runs/24353750976/job/71115005146 

### Checklist


.github/workflows/terraform_modules_test.yaml
```diff
@@ -9,6 +9,9 @@ name: Terraform modules tests
 on:
   workflow_call:
     inputs:
+      additional-setup-script:
```
swetha1654, line 12:
```suggestion
      additional-terraform-setup-script:
```
to be more explicit

.github/workflows/terraform_modules_test.yaml
```diff
@@ -88,21 +91,32 @@ jobs:
...
         sudo k8s config | juju add-k8s tfk8s --client
-        juju bootstrap tfk8s
-        juju add-model tf-testing-k8s --controller tfk8s
-    - name: Prepare environment for Terraform tests with lxd controller
-      if: ${{ inputs.lxd-controller }}
-      run: |
-        sudo snap install lxd
-        juju bootstrap localhost lxd
-        juju add-model tf-testing-lxd --controller lxd
+        if [ "$LXD_ENABLED" == "true" ]; then
```
javierdelapuente: So as input variables we have for this workflow:
k8s-controller
lxd-controller
k8s-channel

If we put 
k8s-controller True and lxd-controller False, it creates an k8s-controller
k8s-controller False and lxd-controller True, it creates an lxd-controller
k8s-controller True and lxd-controller True, it creates an lxd-controller and puts the k8s cloud in the lxd controller.

I think the naming of the variables is not confusing and requires a refactor.

seb4stien: Would you have a proposal?
Most of the time we will fall into a correct name [k8s|lxd]-controller. So "most" of the time the naming is correct :)

Another option could be to forget about the booleans and go for something like "juju-deployment: [k8s-only|lxd-only|lxd+k8s-cloud]", WDYT?

.github/workflows/terraform_modules_test.yaml
seb4stien: Not sure. It can actually be anything that has to be run before the terraform steps.

.github/workflows/terraform_modules_test.yaml
javierdelapuente: I mean, if k8s-controller True and lxd-controller True, I would expect two controllers to be created (that could be a valid use case, using cross controller relations). The description for `k8s-controller` is `Whether to use a Kubernetes controller for testing.`, but if we have the lxd-controller, it is a different thing.

Yeah, I think the non booleans is a good approach.

seb4stien: After thinking about it it would have been a breaking change that would have required to update all repositories. So I introduced a third boolean, and there is a test to be sure that only one of the options is selected. WDYT?

javierdelapuente (approved):

PR #998, opened 2026-04-14
fix: explicitly declare CHARMHUB_TOKEN in workflow_call

ximilian:
### Overview

This PR modifies the promote_charm.yaml reusable workflow to explicitly declare `CHARMHUB_TOKEN` as a required secret in the `workflow_call` trigger.

### Rationale

Currently, this workflow uses `${{ secrets.CHARMHUB_TOKEN }}` within its jobs but does not declare it in the `on: workflow_call: secrets:` block. This creates two issues:
1. **Implicit Dependency:** Calling workflows are forced to use `secrets: inherit` because GitHub Actions prevents passing secrets explicitly to a workflow that hasn't "asked" for them.
2. **Validation:** Users attempting to pass the secret explicitly (e.g., for better security or because their local secret name differs) encounter a workflow validation error.

### Workflow Changes

- Added `secrets: CHARMHUB_TOKEN` to the `workflow_call` configuration.
- Marked the secret as `required: true` since the `promote-charm` job cannot function without it.

- **Backward Compatible:** Users currently using `secrets: inherit` will see no change in behavior; their workflows will continue to work.
- **Improved Security:** Users can now opt to pass only the specific secret required rather than inheriting the entire repository secret context.

### Checklist

- [ ] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied


weiiwang01 (approved):

.github/workflows/promote_charm.yaml
```diff
@@ -45,6 +45,10 @@ on:
         description: Charmcraft channel to use for promoting the charm.
         type: string
         default: latest/stable
+    secrets:
+      CHARMHUB_TOKEN:
+        required: true
```
shan-96, line 50: making it required will break workflows of repos that do not explicitly provide it.

```suggestion
```

ximilian: but making not required will wast CI time because it is needed

Osama-Kassem: AFAIU, `CHARMHUB_TOKEN` was always required, it was just never documented as such. Explicitly documenting the secret and marking it as required will not break existing workflows using `secrets: inherit`.

Osama-Kassem (approved):

shan-96 (approved):

seb4stien (approved):

javierdelapuente (approved):

PR #1000, opened 2026-04-15
Avoid to trigger a charm publication on terraform updates

seb4stien:
Applicable spec: n/a

### Overview

Exclude the "^terraform/*" files from the CODE_FILE_CHANGES count. This way, a PR only changing terraform files (such as a renovate update of dependencies) would not trigger a charm publication to charmhub, which would result in an infinite "renovate" loop (as the tests will be updated with the latest version available on charmhub).

### Rationale

Currently:
- renovate detects a new version of charm on charmhub
- it creates a PR to update the versions in terraform/tests
- the PR is auto-approved
- the PR is merged
- a new version is published on charmhub
- renovate detects a new version of charm on charmhub...

### Workflow Changes

We already had a feature to avoid publishing a charm when there are only documentation updates. I adapted the function to also exclude "^terraform*" file changes.

### Checklist


arturo-seijas (approved):
Note that this implied that a new tag won't be created iirc

PR #1002, opened 2026-04-16
fix: guard against pip break sys package flag

yanksyoon:
Applicable spec: <link>

### Overview

- Adds --brea-system-packages flag for installing pipx

### Rationale

- pip installation requires --break-system-packages flag for newer OSes

### Checklist


.github/workflows/integration_test.yaml
```diff
@@ -229,8 +229,12 @@ jobs:
             sudo apt-get install -y python3-venv
           fi
           if ! command -v pipx &> /dev/null; then
-            python3 -m pip install --upgrade pip
-            python3 -m pip install --user pipx
+            PIP_BREAK_FLAG='' # Guard against older versions of pip that don't have --break-system-packages
```
seb4stien: Isn't it an option to `apt install pipx` instead?

seb4stien (commented):
Shouldn't we `apt install pipx`?

.github/workflows/integration_test.yaml
yanksyoon: Unfortunately, there is no `pipx` available in Jammy through apt. I'll install it via `apt` if it is available.

cbartz (approved):

seb4stien (approved):

PR #1019, opened 2026-06-15
chore: Add steps to sync Vale packages

erinecon:
### Overview

Update `docs.yaml` to download Vale and manually sync the packages.

I used Copilot to prepare this PR.

### Rationale

The workflow started failing in downstream repositories, e.g., https://github.com/canonical/tmate-ssh-server-operator/actions/runs/27526128259/job/81359939373?pr=46

The issue was that there are several styles that this workflow pulls from -- `local`, `PFE`, and `Canonical`. The `vale-action` step performs `vale sync`, but because one of the styles is nested in another (`PFE` nests `Canonical`), the step doesn't capture the Canonical style in its entirety, thus leading to a `dicpath` error. 

### Workflow Changes

Add two new steps to `docs.yaml`:
1. Install Vale 
2. `vale sync` to sync all the styles and packages needed

Furthermore, the version of `vale-action` has been bumped and `version: none` has been added to use the previously installed version of Vale.

### Checklist


seb4stien (approved):

.github/workflows/docs.yaml
```diff
@@ -47,12 +47,21 @@ jobs:
         run: |
           sudo apt-get update -y
           sudo apt-get install -y docutils
+      - name: Install Vale
+        run: |
+          gh release download --repo errata-ai/vale --pattern 'vale_*_Linux_64-bit.tar.gz' --output vale.tar.gz
```
javierdelapuente: Why not the snap, it looks it is the recommendad way to install it: https://vale.sh/docs/install

.github/workflows/docs.yaml
```diff
@@ -47,12 +47,21 @@ jobs:
...
+      - name: Sync Vale packages
+        run: vale sync
       - name: Run Vale Linter
-        uses: vale-cli/vale-action@v2.1.1
+        uses: vale-cli/vale-action@2.1.2
         with:
           files: ${{ inputs.vale-files }}
           vale_flags: ${{ inputs.vale-flags }}
           fail_on_error: ${{ inputs.vale-fail-on-error }}
+          version: none
```
javierdelapuente, line 60: Not sure I understood why vale was not installed with this action?

Following https://github.com/vale-cli/vale-action#version-default-latest by default latest would be installed

erinecon: Vale was installed with the action. But now that I added a previous step to install Vale + sync the packages, we need `vale-action` to use that previously installed version instead of installing a brand new version

.github/workflows/docs.yaml
erinecon: I actually wanted to use the snap but Copilot advised against it:

> ...it's not recommended here. `snap` is available on `ubuntu-latest` runners but has a significant drawback in CI: snap installations are asynchronous. The snap daemon installs packages in the background, so immediately running `vale sync` after `sudo snap install vale` can fail because the `vale` binary isn't on the PATH yet. You'd need a workaround like polling for the binary.

WDYT?

.github/workflows/docs.yaml
```diff
@@ -47,12 +47,21 @@ jobs:
...
           sudo apt-get update -y
           sudo apt-get install -y docutils
+      - name: Install Vale
+        run: |
+          gh release download --repo errata-ai/vale --pattern 'vale_*_Linux_64-bit.tar.gz' --output vale.tar.gz
+          tar -xz -C /usr/local/bin -f vale.tar.gz vale
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Sync Vale packages
+        run: vale sync
```
srbouffard, line 53: I am a bit surprised that we have to be explicit with `vale sync` and that fixes the issue. 
Both the explicit call and the action perform the same thing behind the scenes.

Image

So I wonder if there is something introduced in 2.1.2 that would explain why now it's failing.

.github/workflows/docs.yaml
srbouffard: We should update `errata-ai/vale` to point to their new name: https://github.com/vale-cli/vale

.github/workflows/docs.yaml
erinecon: Based on my debugging session with Copilot, the issue seems to be that, while `vale-action` does perform a sync, that sync does not pick up on the nested styles hosted in a separate repository, and that is what throws the dicpath error. So either we need to commit the styles in this repo, or we need to sync all the styles before `vale-action` is run.

And Copilot also pointed to a recent release of Vale ([3.15.0](https://github.com/vale-cli/vale/releases/tag/v3.15.0)) where bug fixes and updates were made to `vale sync` that directly relate to the failures we're experiencing. So there is a recent change that aligns with the failures we're experiencing.

.github/workflows/docs.yaml
erinecon: Great catch! Depending on how we wrap up the conversation around this step (using `snap install` or not), I'll update this line :D

javierdelapuente: We use snaps everywhere! If it does not work, then the snap is not correct.

erinecon: Yeah, I was wondering if Copilot was being overly cautious :sweat_smile: I'll update the step, thanks!

erinecon: Opting to use `snap install` for this step instead, but thanks again for catching the old name!

PR #1027, opened 2026-07-09
feat: Add lxd-channel input for custom LXD setup in integration tests

HomayoonAlimohammadi:
This pull request adds support for specifying a custom LXD snap channel in the integration test GitHub Actions workflow. This allows users to control which LXD version is installed during CI runs, improving flexibility and reproducibility.

**Workflow input enhancements:**

* Added a new `lxd-channel` input to `.github/workflows/integration_test.yaml`, allowing the snap channel for LXD installation to be specified. If left empty, the default channel from the setup-lxd action is used.

**Conditional LXD setup steps:**

* Updated the integration test job to set up LXD using either the default or a custom snap channel, based on the `lxd-channel` input. This is achieved by conditionally running the `canonical/setup-lxd` action with or without the `channel` parameter.

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -38,7 +38,7 @@ on:
         description: Actions operator juju channel as per https://github.com/charmed-kubernetes/actions-operator#usage
         type: string
         default: 2.9/stable
-      lxd-channel:
+      build-lxd-channel:
         description: Snap channel to install LXD from in the build job via canonical/setup-lxd.
         type: string
         default: "latest/candidate"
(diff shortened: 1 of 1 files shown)


.github/workflows/integration_test.yaml
```diff
@@ -38,6 +38,12 @@ on:
         description: Actions operator juju channel as per https://github.com/charmed-kubernetes/actions-operator#usage
         type: string
         default: 2.9/stable
+      lxd-channel:
+        description: >-
+          Snap channel to install LXD from in the build job via canonical/setup-lxd.
+          If left empty, the setup-lxd action's own default channel is used.
+        type: string
+        default: ""
```
javierdelapuente: Could we set the default to `'latest/candidate'? That way we can use only one setup LXD step with the lxd-channel. WDYT?

HomayoonAlimohammadi: Done, thank you!

javierdelapuente (approved):

HomayoonAlimohammadi:
I've force pushed to 1. remove the noisy revert commit, and 2. to push a verified commit (CI was blocked due to unverified commits).

PR #1030, opened 2026-08-11
 Search for charmcraft files alias before lib-check 

weiiwang01:
Applicable spec: <link>

### Overview

The `lib-check` job runs `canonical/charming-actions/check-libraries`, which expects a `charmcraft.yaml` at the configured charm path. But now some repositories will use a prefixed `charmcraft.yaml` (e.g. [`chrony-charmcraft.yaml`](https://github.com/canonical/chrony-operators/pull/16)) which will cause the job to fail.

### Rationale

Support repositories that use a prefixed `charmcraft.yaml` (e.g. [`chrony-charmcraft.yaml`](https://github.com/canonical/chrony-operators/pull/16)).

### Workflow Changes

The `lib-check` job now moves a prefixed `*charmcraft.yaml` or `*charmcraft.yml` file to `charmcraft.yaml` when neither `charmcraft.yaml` nor `charmcraft.yml` exists.

### Checklist

### DIFF
--- a/.github/workflows/auto_update_charm_libs.yaml
+++ b/.github/workflows/auto_update_charm_libs.yaml
@@ -41,9 +41,9 @@ jobs:
           if [ ${#candidates[@]} -eq 0 ]; then
             exit 0
           fi
-          echo "Copying ${candidates[0]} to charmcraft.yaml"
-          cp "${candidates[0]}" charmcraft.yaml
-          echo "moved=true" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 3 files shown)


yanksyoon (approved):
LGTM, would symlinking rather than moving it also be something viable? This seems more semantically meaning "temporary file" than copying.
Thanks @weiiwang01 !

weiiwang01:
> LGTM, would symlinking rather than moving it also be something viable? This seems more semantically meaning "temporary file" than copying. Thanks @weiiwang01 !

That's a good idea, updated, thanks!

PR #1031, opened 2026-08-13
feat: add terraform version tagging workflow

f-atwi:
Applicable spec: [CC008](https://docs.google.com/document/d/1k1psLCfcf0Nr5KeVtWGFi9wmzhcg5AP9GVZTsRyVQSQ/edit?usp=sharing)

### Overview

Adds a reusable workflow (`terraform_modules_release.yaml`) that automatically computes and pushes a semver tag whenever Terraform module changes are merged to a caller's main branch (in compiance with CC008).

Versioning scheme: `<prefix>-<major>.<minor>.0`
- Major is read from a file in the caller's repo (default: `terraform/MAJOR_VERSION`)
- Minor is auto-incremented from the highest existing tag for that major
- Patch is always `0` 

### Rationale

Terraform modules need versioned releases so consumers can pin to a stable ref. This automates that process in the same reusable-workflow pattern already used for charm publishing and testing, avoiding per-repo maintenance.

### Workflow Changes

`terraform_modules_tag.yaml` — new reusable workflow:
- Two overridable inputs: `tag-prefix` (default: `tf`) and `major-version-file` (default: `terraform/MAJOR_VERSION`)
- Concurrency group with `cancel-in-progress: false` prevents two runs from computing the same next minor; the in-progress run always completes before a queued run starts

### Usage

Callers add to their repo:

```yaml
on:
  push:
    branches: [main]
    paths: ['terraform/**']

jobs:
  tag:
    uses: canonical/operator-workflows/.github/workflows/terraform_modules_release.yaml@main
    with:                                        # optional: override  defaults
      tag-prefix: tf 
      major-version-file: terraform/MAJOR_VERSION
```

### Checklist

- [ ] The [contributing guide](https://github.com/canonical/platform-engineering-contributing-guide) was applied

### DIFF
--- a/.github/workflows/terraform_modules_release.yaml
+++ b/.github/workflows/terraform_modules_release.yaml
@@ -65,7 +65,7 @@ jobs:
 
           tag="${TAG_PREFIX}-${major}.${new_minor}.0"
           if git rev-parse -q --verify "refs/tags/${tag}" >/dev/null; then
-            echo "${tag} unexpectedly already exists; refusing to overwrite." >&2
+            echo "::warning:: ${tag} unexpectedly already exists; refusing to overwrite." >&2
             exit 1
           fi
 
--- a/.github/workflows/terraform_modules_release.yaml
+++ b/.github/workflows/terraform_modules_release.yaml
@@ -1,7 +1,8 @@
 # Copyright 2026 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-name: Terraform Modules Tag
+name: Terraform Modules Release
+
 on:
   workflow_call:
     inputs:
--- a/.github/workflows/terraform_modules_release.yaml
+++ b/.github/workflows/terraform_modules_release.yaml
@@ -70,5 +70,5 @@ jobs:
           fi
 
           echo "Creating ${tag}"
-          git tag "${tag}"
-          git push origin "${tag}"
+          git tag -- "${tag}"
+          git push origin -- "${tag}"
--- a/.github/workflows/terraform_modules_release.yaml
+++ b/.github/workflows/terraform_modules_release.yaml
@@ -8,8 +8,8 @@ on:
     inputs:
       tag-prefix:
         type: string
-        description: Prefix used for the version tag.
-        default: tf
+        description: "Tag prefix including separator; resulting format is <prefix><major>.<minor>.0"
+        default: tf-
       major-version-file:
         type: string
         description: Path to the file containing the major version number.
@@ -24,7 +24,7 @@ concurrency:
   cancel-in-progress: false
 
 env:
-  TAG_PREFIX: ${{ inputs.tag-prefix || 'tf' }}
+  TAG_PREFIX: ${{ inputs.tag-prefix || 'tf-' }}
   MAJOR_VERSION_FILE: ${{ inputs.major-version-file || 'terraform/MAJOR_VERSION' }}
 
 jobs:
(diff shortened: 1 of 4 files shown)


.github/workflows/terraform_modules_release.yaml
```diff
@@ -0,0 +1,74 @@
...
+permissions:
+  contents: read
+
+# Serialize tag creation so concurrent pushes don't compute the same minor.
+concurrency:
+  group: tag-terraform-module
+  cancel-in-progress: false
+
+env:
+  TAG_PREFIX: ${{ inputs.tag-prefix || 'tf' }}
```
yanksyoon: Should the default tag prefix be `tf-`?

alithethird (approved):

.github/workflows/terraform_modules_release.yaml
```diff
@@ -0,0 +1,74 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Compute and push the next Terraform module tag
+        run: |
+          set -euo pipefail
+
+          # MAJOR_VERSION holds only the major version. Minor is auto-bumped
+          # on every change under terraform/; patch is always 0.
+          major="$(grep -oE '^[0-9]+' "${MAJOR_VERSION_FILE}" | head -1)"
```
yanksyoon, line 46: Is this our new team standard? It seems that it may start failing workflows on repositories with no terraform major version file set.

.github/workflows/terraform_modules_release.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+          tag="${TAG_PREFIX}-${major}.${new_minor}.0"
+          if git rev-parse -q --verify "refs/tags/${tag}" >/dev/null; then
+            echo "${tag} unexpectedly already exists; refusing to overwrite." >&2
+            exit 1
+          fi
+
+          echo "Creating ${tag}"
+          git tag -- "${tag}"
+          git push origin -- "${tag}"
```
yanksyoon, line 74: Would this require any additional permissions? Would it help for the workflow to explicitly define what permissions it needs?

.github/workflows/terraform_modules_release.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+          if [ -z "${latest_minor}" ]; then
+            new_minor=0
+          else
+            new_minor=$((latest_minor + 1))
+          fi
+
+          tag="${TAG_PREFIX}-${major}.${new_minor}.0"
+          if git rev-parse -q --verify "refs/tags/${tag}" >/dev/null; then
+            echo "${tag} unexpectedly already exists; refusing to overwrite." >&2
```
yanksyoon:
```suggestion
            echo "::warning:: ${tag} unexpectedly already exists; refusing to overwrite." >&2
```
Nit: Pretty formatting echos might make debugging easier!
Ref: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands

yanksyoon (approved):
Minor nits, thanks @f-atwi !

seb4stien:
Can you please describe the workflow in README.md?

.github/workflows/terraform_modules_release.yaml
f-atwi: the `-` is added automatically making the format <prefix>-X.Y.Z
however, I think the logic should be changed so that the `-` must be explicitly added in here which would allow us to omit it if we ever decide on a different tag format
will change it

.github/workflows/terraform_modules_release.yaml
f-atwi: Yes. In compliance with CC008 spec, we need to be adding tags for terraform module releases.
The decision as discussing in an EMEA design meeting is to keep patch version at 0, only incrementing the minor, and have a way to identify what version is the major.

Though this means we need to add a major version file on all repos, at least we do not have to do any manual changes for versioning when modifiying the terraform files (unless we need to change the major version :) )

If you have any counter proposals of how to do this that would be great. Best to discuss now before starting to use this everywhere

.github/workflows/terraform_modules_release.yaml
f-atwi: Unless I misunderstood you, this workflow already mentions `permissions: contents: write` which is what we need here. All workflows using this file would inherit the permissions (or use `permissions: inherit` explicitly)

PR #1044, opened 2026-08-27
chore: remove usage of charming-actions in promote

javierdelapuente:
Applicable spec: <link>

### Overview

Remove charming-actions from the promote workflow. Instead, use plain bash and charmcraft commands.

The main reason for this change is that the release is coupled to the upload in charming-actions, and requires a release with a certain name. We do not want to have one release per revision (as for a charm with many architectures, bases...) that will mean lots of releases and noise. That functionality was removed from charm-ci (and instead it creates a tag per revision and only one release for all of them), see https://github.com/canonical/charm-ci/pull/129

Example working execution: https://github.com/javierdelapuente/javi-traefik-k8s-operator/actions/runs/33060646294/job/98478286732

### Checklist

### DIFF
--- a/.github/workflows/promote_charm.yaml
+++ b/.github/workflows/promote_charm.yaml
@@ -58,9 +58,9 @@ jobs:
       CHARMCRAFT_AUTH: ${{ secrets.CHARMHUB_TOKEN }}
     outputs:
       charm-name: ${{ steps.set-outputs.outputs.charm-name }}
-      name: ${{ steps.set-outputs.outputs.name }}
-      channel: ${{ steps.set-outputs.outputs.channel }}
-      architecture: ${{ steps.set-outputs.outputs.architecture }}
+      base-name: ${{ steps.set-outputs.outputs.base-name }}
+      base-channel: ${{ steps.set-outputs.outputs.base-channel }}
+      base-architecture: ${{ steps.set-outputs.outputs.base-architecture }}
       revision: ${{ steps.set-outputs.outputs.revision }}
       resource-flags: ${{ steps.set-outputs.outputs.resource-flags }}
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/promote_charm.yaml
```diff
@@ -36,7 +36,7 @@ on:
         type: string
         required: false
         description: |
-          Tag prefix, useful when promoting multiple charms from the same repo.
+          Deprecated and unused. Kept for backwards compatibility with existing callers.
```
swetha1654, line 39: i didn't realise this option existed. Won't using this solve our problem on traefik?

.github/workflows/promote_charm.yaml
```diff
@@ -83,84 +88,119 @@ jobs:
...
+          set -e
+
+          if [[ -z "$ORIGIN_CHANNEL" ]]; then
+            echo "::error::Input 'origin-channel' is required and was not supplied"
+            exit 1
+          fi
+          if [[ -z "$DESTINATION_CHANNEL" ]]; then
+            echo "::error::Input 'destination-channel' is required and was not supplied"
+            exit 1
+          fi
```
swetha1654, line 109: we could also check for whether origin != destination and !(origin == 'stable' and destination == 'edge') perhaps?

swetha1654 (approved):

.github/workflows/promote_charm.yaml
javierdelapuente: Not really, it looks for a release, not a tag (I broke that on purpose to not have 20 releases of the same code, one for revision).

.github/workflows/promote_charm.yaml
```diff
@@ -57,15 +57,20 @@ jobs:
     env:
       CHARMCRAFT_AUTH: ${{ secrets.CHARMHUB_TOKEN }}
     outputs:
-      name: ${{ steps.set-base-outputs.outputs.name }}
-      channel: ${{ steps.set-base-outputs.outputs.channel }}
-      architecture: ${{ steps.set-base-outputs.outputs.architecture }}
+      charm-name: ${{ steps.set-outputs.outputs.charm-name }}
+      name: ${{ steps.set-outputs.outputs.name }}
```
seb4stien: If that's purely internal, can't we explicitely name it base-name?

seb4stien (approved):

.github/workflows/promote_charm.yaml
javierdelapuente: Done the not equal thing. The other is pretty complex (a track is can also have a branch), so I skipped it.

PR #1045, opened 2026-08-31
feat: add terraform compliance workflow 

f-atwi:
Applicable spec: [CC008](https://docs.google.com/document/d/1k1psLCfcf0Nr5KeVtWGFi9wmzhcg5AP9GVZTsRyVQSQ/edit?usp=sharing)

### Overview

Adds a terraform compliance workflow to ensure that the terraform modules align with the CC008 spec

An example of this workflow running on a PR:
https://github.com/canonical/gateway-api-integrator-operator/pull/343

### Rationale

Guards changes to be pushed from drifting away from the spec

### Checklist

- [ ] The [contributing guide](https://github.com/canonical/platform-engineering-contributing-guide) was applied


.github/workflows/terraform_modules_compliance.yaml
```diff
@@ -0,0 +1,43 @@
...
+    name: CC008 compliance
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout caller repository
+        uses: actions/checkout@v7.0.1
+      - name: Checkout compliance checker
+        uses: actions/checkout@v7.0.1
+        with:
+          repository: canonical/operator-workflows
+          ref: feat/terraform-compliance
```
javierdelapuente: I would use `canonical/get-workflow-version-action` instead

See https://github.com/canonical/charm-ci/blob/c3353e07847fa7a1f9d7806690382cd78999ce33/.github/workflows/integration-test.yml#L129 for an example

seb4stien:
@f-atwi could you link an real run in the PR description? (= a repo using the workflow in the branch)

f-atwi:
> @f-atwi could you link an real run in the PR description? (= a repo using the workflow in the branch)

@seb4stien done

terraform-compliance/README.md
```diff
@@ -0,0 +1,59 @@
...
+Module type is inferred: no `module` blocks → charm; composes modules and
+defines a `juju_model`/`juju_secret`/`juju_integration`/`juju_offer` →
+product; composes only → component.
+
+## Known deviations
+
+- `providers.tf` is not required (CC008 modules are non-root, so have no
+  provider config to place there).
+- Output types aren't checked (Terraform infers them from the value).
+- `provides`/`requires` outputs are optional: CC008 makes them mandatory only
```
seb4stien, line 59: Could/Should we have some test if these variables are present? I would also return a violation if some "old" variables are present in the output (e.g. "endpoint" / "endpoints" IIRC).

seb4stien (approved):

terraform-compliance/README.md
f-atwi: we are checking the existence of them but not the type
and we are not failing if optional ones are not there
i have created a deprecated names that for now only holds "endpoint" and "endpoint" to make sure they do not exist
it would be hard to enforce checking that there are no variable names outside what is defined here as required and optional since even the spec itself says it permits additional variables

seb4stien: Checking for "endpoint" and "endpoints" is good for me. Thanks.

terraform-compliance/terraform_check.py
```diff
@@ -0,0 +1,441 @@
+# Copyright 2026 Canonical Ltd.
```
seb4stien, line 1: To be renamed "terrafor_check.py" and the following one "terraform_spec.py" so that we can evolve them over time.

terraform-compliance/cc008_spec.py
```diff
@@ -0,0 +1,170 @@
...
+                # Optional: subordinate charms must omit units, and
+                # subordinate-ness isn't visible from Terraform.
+                VariableRule("units", TypeFamily.NUMBER, default=1, optional=True),
+                # Optional CC008 charm variables (validated only when present).
+                VariableRule("base", TypeFamily.STRING, default=None, optional=True),
+                VariableRule("expose", TypeFamily.COLLECTION, default={}, optional=True),
+                VariableRule("resources", TypeFamily.COLLECTION, default={}, optional=True),
+                VariableRule("machines", TypeFamily.COLLECTION, default=[], optional=True),
+                VariableRule(
+                    "endpoint_bindings", TypeFamily.COLLECTION, default={}, optional=True
```
Ignavar: Per the [Terraform Registry docs](https://registry.terraform.io/providers/juju/juju/latest/docs/resources/application#endpoint_bindings-1) ,  endpoint_bindings  is a  set(object({ space, endpoint })) , not a map. Terraform rejects  default = {}  for this type the only valid empty default is  [] . Could  cc008_spec.py 's  VariableRule  for  endpoint_bindings  be updated to expect  []  instead of  {} ?

